PR #1, opened 2021-04-17
Use Quilt Json5 (but always strict)

TheGlitch76:


build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
 	implementation "org.ow2.asm:asm-tree:${project.asm_version}"
 	implementation "org.ow2.asm:asm-util:${project.asm_version}"
 
+	implementation 'org.quiltmc:quilt-json5:1.0.0-rc.3' // TODO: shadow me
```
liach, line 42: You can use a custom built johnrengelman shadow plugin for now, I assume?

i509VCB: Yeah that is a pain point - loom and shadow are not friends. Impl detail for now and I do plan to shadow it later.

zml2008: loader should be fine -- loom will be its own adventure, have to disable gradle's automatic publishing and replace it wholesale to be able to use shadow

PR #2, opened 2021-04-21
Re-implement process forking to make the loader error gui work on mac os.

AlexIIL:
This a feature that I originally made for fabric-loader PR 154, but was removed for some reason that I can't quite find. This used to be fabric-loader PR 341

This doesn't handle input from the forked process - which could be used in the future for custom buttons (such as "open mod directory" or "change a config value"). However as fabric-loader doesn't add any yet I'm not sure if that's a problem yet.

Currently the `QuiltStatusTree` json reading expects everything in a specific order - as this is currently only written in one place (by the "write" methods) and stored in a temporary file, I don't expect this to be an issue. If needed though I can change that to a proper loop and switch statement?

NOTE: I haven't tested quilt loader *at all* yet, although I will soon. This is mostly just opening after porting.


AlexIIL:
Oh, and the `quilt_x16` and `quilt_8` pngs will need replacing with the quilt logo, rather than the fabric one.

src/main/java/org/quiltmc/loader/impl/gui/QuiltGuiEntry.java
```diff
@@ -109,4 +161,74 @@ private static void addThrowable(QuiltStatusNode node, Throwable e, Set<Throwabl
 			addThrowable(sub, t, seen);
 		}
 	}
+
+	private static void fork(QuiltStatusTree tree, boolean shouldWait) throws Exception {
+		List<String> commands = new ArrayList<>();
+		commands.add(System.getProperty("java.home") + File.separator + "bin" + File.separator + "java");
```
i509VCB, line 167: How does vanilla launcher handle this? On windows at least the launcher will use the java executable in launcher folder, so this may fail if on macOS forking the JVM does not have a path set BUT a different jvm launched it.

AlexIIL: Why would that affect this?

According to `java.lang.System.getProperties()`:
> java.home | Java installation directory

Which, to me, indicates we'd always fork with whatever JVM was used to run minecraft. (As for how jvms are organised... I have no idea, I'm really just assuming that this `<java-home>/bin/java` is always valid)

i509VCB: Ah didn't realize that.

src/main/java/org/quiltmc/loader/impl/gui/QuiltStatusTree.java
```diff
@@ -64,15 +92,24 @@ public static FabricTreeWarningLevel getHighest(FabricTreeWarningLevel a, Fabric
...
+
 	/** Generic JSON file. */
 	public static final String ICON_TYPE_JSON = "json";
 
 	/** A file called "fabric.mod.json". */
 	public static final String ICON_TYPE_FABRIC_JSON = "json+fabric";
 
+    /** A file called "quilt.mod.json". */
+    public static final String ICON_TYPE_QUILT_JSON = "json+quilt";
+
```
Earthcomputer, line 112: Should use tabs rather than spaces

AlexIIL: Oh, right, I haven't setup quilt to use tabs for me yet - thanks :p

i509VCB:
Content looks good - would like to test this before I consider merging this.

AlexIIL:
I tested this a few days ago, and it seemed to work fine.
image

(The logo displays when the window is minimised, but sadly not when it is shown - in that case it just displays the "java logo" in the dock. However that's a minor bug that can probably be looked into later).

src/main/java/org/quiltmc/loader/impl/gui/QuiltGuiEntry.java
```diff
@@ -109,4 +161,74 @@ private static void addThrowable(QuiltStatusNode node, Throwable e, Set<Throwabl
...
+	/** The entry point after forking the main application over into a different process to get around incompatibilities
+	 * on OSX (and problems where some games switch swing over to use headless mode, which doesn't work very well for
+	 * us). */
+	public static void main(String[] args) throws Exception {
+		if (args.length == 2 && "--read-tree".equals(args[0])) {
+			System.out.println("Status:Reading");
+			File from = new File(args[1]);
+
+			try (InputStream is = new GZIPInputStream(new BufferedInputStream(new FileInputStream(from)))) {
+				JsonReader reader = JsonReader.create(new InputStreamReader(is, StandardCharsets.UTF_8));
```
TheGlitch76, line 218: is creating a json5 reader intentional here? I probably need to change these names one more time...

AlexIIL: Am I meant to use `new JsonReader(Reader)` instead? For some reason I don't have docs for json5 in my IDE.

Or am I not meant to be using `quilt-json5` for this at all? I'm currently assuming we need to keep loader far away from any dependencies, which I assumed we want to use json5 for json handling?

(Either way, what I'm trying to do here is read and write a json file automatically, that the user isn't expected to see normally. For the sake of debugging it's exported with a single space for indenting, but there's definitely no need for comments).

TheGlitch76: Use .createStrict here

i509VCB: I think we should flip the options glitch, this is a confusion I see quite often

AlexIIL: ok

TheGlitch76: See the issue I opened on the github for it to pitch in name change ideas. i know it needs to change before it's stabilized but idk to what

i509VCB:
Manually merged via 2d98f378371bb26a8807b2ea40925c04faf5feed

PR #6, opened 2021-04-21
Ignore $schema in v1 fabric.mod.json files instead of warning

kb-1000:
[JSON Schema](https://json-schema.org) specifies that the `$schema` key is used to set the JSON schema used for the file.
By ignoring this key instead of printing a warning, it becomes possible to specify such a schema for the editor support _without_ causing warnings on every load of the mod.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/metadata/ModMetadataParser.java
+++ b/src/main/java/org/quiltmc/loader/impl/metadata/ModMetadataParser.java
@@ -65,7 +65,8 @@ public final class ModMetadataParser {
 
 				while (reader.hasNext()) {
 					// Try to read the schemaVersion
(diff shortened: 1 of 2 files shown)


i509VCB:
This should be good, however I would like you to add the additional check to avoid the slower route scan for schema version given `$schema` as the first field in an FML about hereish: https://github.com/QuiltMC/quilt-loader/blob/master/src/main/java/org/quiltmc/loader/impl/metadata/ModMetadataParser.java#L68

i509VCB (approved):

i509VCB:
Added a slight formatting change, going to merge this.

i509VCB:
Darn forgot to squash - time to set the rules on repo

PR #9, opened 2021-04-21
Clean up compiler warnings

Nuclearfarts:
There are quite a few compiler warnings, most of them trivial things like rawtypes, unneeded `@SuppressWarnings` annotations, unchecked casts, and missing `serialVersionUID`. This fixes the rawtypes, properly suppresses the unchecked casts rather than only suppressing them in IDEA, suppresses the `serialVersionUID` complaints since we don't use `Serializable` anyway, removes some unused private methods and constructors, and moves away from using deprecated code where possible and suppresses the deprecation warning where it makes sense to do so (usually for things that are either for support of deprecated things or when the `@Deprecated` is because it's internal rather than because it's deprecated.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ plugins {
 	id 'eclipse'
 	id 'maven-publish'
 	id("org.cadixdev.licenser") version "0.5.0"
-    id("org.quiltmc.loom") version "0.7.13"
+	id("org.quiltmc.loom") version "0.7.13"
 }
 
 sourceCompatibility = 1.8
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ plugins {
 	id 'eclipse'
 	id 'maven-publish'
 	id("org.cadixdev.licenser") version "0.5.0"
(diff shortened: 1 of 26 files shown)


build.gradle
```diff
@@ -4,7 +4,7 @@ plugins {
 	id 'eclipse'
 	id 'maven-publish'
 	id("org.cadixdev.licenser") version "0.5.0"
-    id("org.quiltmc.loom") version "0.7.12"
+    id("org.quiltmc.loom") version "0.7.13"
```
i509VCB, line 7: Could you make sure you set the buildscript to indent with tabs in eclipse?

i509VCB (approved):
Everything else looks good aside from one line indentation in buildscript.

build.gradle
Nuclearfarts: What the heck? I have eclipse set to use tabs everywhere as it's my preference as well. Looking at the diff who knows how long that's been spaces. Fixed it anyway.

TheGlitch76 (approved):

PR #10, opened 2021-04-22
Swap out Fabric logos for Quilt ones

TheEssem:
Simple change, but it had to be done eventually.


i509VCB:
I think the assets may be referred to in spots in the code - could you make sure those get changed?

Prospector:
Also, for any small icons, make sure to use the "mini icon" variant of the logo, it's more optimized for small sizes.

TheEssem:
> I think the assets may be referred to in spots in the code - could you make sure those get changed?

None of the filenames were changed and the assets seem to be working fine here, so I think it should be fine.

> Also, for any small icons, make sure to use the "mini icon" variant of the logo, it's more optimized for small sizes.

Fixed.

i509VCB:
I think the fabric.mod.json may have one?

i509VCB (approved):

misode:
It seems like #2 already made these changes https://github.com/QuiltMC/quilt-loader/pull/2/commits/e8726d2c4843680b545f06af7c9960132c547a11#diff-24b3f3253a483300eaffcab8a5f4462f6bb4e6b42bb12997a3efa0d9aa22d86b

TheEssem:
> It seems like #2 already made these changes [e8726d2#diff-24b3f3253a483300eaffcab8a5f4462f6bb4e6b42bb12997a3efa0d9aa22d86b](https://github.com/QuiltMC/quilt-loader/commit/e8726d2c4843680b545f06af7c9960132c547a11#diff-24b3f3253a483300eaffcab8a5f4462f6bb4e6b42bb12997a3efa0d9aa22d86b)

In that case, if that PR gets merged then I'll remove the other icons and keep only the mod icon.

AlexIIL:
`/ui/icon/decoration/fabric_8x` and `/ui/icon/fabric_16x` will eventually be used to indicate *fabric* mods in the error gui, rather than quilt mods, and as such they should be kept around until we completely remove all possibility of loading fabric mods in quilt. Instead, can you add the quilt icons as `/ui/icon/decoration/quilt_8x` and `/ui/icon/quilt_16x`?

`/ui/icon/fabric_128x` should be renamed to `/ui/icon/quilt_128x` though, and the line loading it should be changed (https://github.com/QuiltMC/quilt-loader/blob/master/src/main/java/org/quiltmc/loader/impl/gui/QuiltMainWindow.java#L98). However, this is used for the window and taskbar icon, and as such it should probably use a blown-up version of the smaller one (or use `window.setIconImages()` but passed a list of `quilt_8x` in powers for however many we want).

AlexIIL:
Actually, I should probably correct myself: the images in `/src/main/resources/ui/icons` were intended to be used to display both the error gui that we have now, but also a "filesystem info" tree and a "mods list" from https://github.com/FabricMC/fabric-loader/pull/154. However the last two tabs weren't wanted, and I'm not sure if quilt wants those to be brought back. (Either way, if they are wanted, then it will probably make sense to correctly identify fabric mods from quilt mods. However this is minor either way - unlike  `src/main/resources/assets/quilt_loader/icon.png` which absolutely should be updated and I have no issues with being quilt rather than fabric).

AlexIIL:
Thanks! It looks good now :)

*Mockup:*
Screenshot from 2021-04-22 20-45-48

AlexIIL (approved):

Prospector (approved):

Earthcomputer (approved):

PR #12, opened 2021-04-22
[PROOF_OF_CONCEPT] Remote mod support for mod resolution

AlexIIL:
Unfinished.

Uses mod resolver functionality added in https://github.com/QuiltMC/quilt-loader/pull/11 to choose the best version(s) of libraries to download, and only the ones which work with your existing mod set.

To test this you'll need to create a mod jar with this for it's `fabric.mod.json` content:
````json
{
    "schemaVersion": 1,
    "id": "netherite_plus",
    "name": "Netherite Plus Mod",
    "version": "1.0.0",
    "depends": {
        "netherite-tools-lib": "*"
    }
}
````

This functions by fetching the remote file https://alexiil.uk/quilt_temp_maven/index/netherite-tools-lib.json.gz, which contains this content:
````json
{
    "id": "netherite-tools-lib",
    "versions": {
        "1.0.0": {
            "depends": {
                "minecraft": "=1.16.4"
            }
        },
        "1.1.0": {
            "depends": {
                "minecraft": "=1.16.4"
            }
        },
        "1.2.0": {
            "depends": {
                "minecraft": "=1.16.5"
            }
        }
    }
}
````
Essentially, this is a stripped-down mod json file (since most of the information isn't needed - only resolution information).

The result of this is that version "1.1.0" will be picked, since 1.2.0 is incompatible with minecraft 1.16.4 (the version that loader is based on), and since 1.0.0 is an older version it is *not* picked.

This way of handling mod downloading should help simplify depending on an external library, since you don't need to directly link to a single version. (Admittedly I don't know exactly how you would have wanted to handle dep downloading, but doing it alongside the rest of resolution is probably better than just downloading single jars).

This PR has huge unfinished areas - for this to actually be useful we'd need to download jar files, store downloaded jars in a separate folder (maybe in `<mc-dir>/.quilt/jars/<mod-name>/mod_file.jar`?). Ideally we'd also cache the downloaded index file in a global folder rather than downloading every index every time. In addition this doesn't handle any of the more complicated cases of mod dependencies.

For a better diff please look at https://github.com/AlexIIL/quilt-loader/pull/1, since this is based on my previous PR for mod resolution.


AlexIIL:
Since I don't actually provide a valid `ModCandidate` yet this will crash when trying to load the mod :p

Also, actually trying to launch results in a "crashed-but-still-running" JVM that needs to be manually stopped - it looks like QuiltLoader doesn't handle crashes from tiny-remapper properly, and a "fork-join pool" is still alive.

AlexIIL:
Closing since this doesn't seem to reflect the (simpler) direction we'd like to take dep downloading.

PR #20, opened 2021-06-06
Allow for customizing the mods and config folder with VM args

FoundationGames:
This PR lets you change the subfolders (within the current game directory) from which mods and mod configuration files are loaded.
The arguments are as follows:
`-Dquilt.modsDir=path/to/mods`
`-Dquilt.configDir=path/to/config`

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -120,27 +120,13 @@ public class QuiltLoaderImpl implements FabricLoader {
 	private void setGameDir(Path gameDir) {
 		this.gameDir = gameDir;
 		String configDir = System.getProperty(SystemProperties.CONFIG_DIRECTORY);
-		this.configDir = resolveIfNotAbsolute(gameDir, configDir == null || configDir.isEmpty()  ? "config" : configDir);
+		this.configDir = gameDir.resolve(configDir == null || configDir.isEmpty()  ? "config" : configDir);
 		refreshModsDir(gameDir);
 	}
 
 	private void refreshModsDir(Path gameDir) {
 		String modsSubDir = System.getProperty(SystemProperties.MODS_DIRECTORY);
-		this.modsDir = resolveIfNotAbsolute(gameDir, modsSubDir == null || modsSubDir.isEmpty() ? "mods" : modsSubDir);
-	}
-
-	/**
-	 * If the passed string toResolve is an absolute path, it will be
-	 * returned as a path. If it is not absolute, it will be returned
-	 * as a subdirectory of the passed rootDir.
-	 *
-	 * @param rootDir The parent directory to use if toResolve is not absolute
-	 * @param toResolve The subdirectory/absolute path to be interpreted
-	 * @return Either a path leading to a subdirectory of rootDir, or an absolute path
-	 */
-	private Path resolveIfNotAbsolute(Path rootDir, String toResolve) {
-		Path abs = Paths.get(toResolve);
-		return abs.isAbsolute() ? abs : rootDir.resolve(toResolve);
+		this.modsDir = gameDir.resolve(modsSubDir == null || modsSubDir.isEmpty() ? "mods" : modsSubDir);
 	}
 
 	@Override
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -55,7 +55,6 @@ import org.quiltmc.loader.impl.launch.knot.Knot;
 import org.quiltmc.loader.impl.metadata.EntrypointMetadata;
 import org.quiltmc.loader.impl.metadata.LoaderModMetadata;
 import org.quiltmc.loader.impl.util.DefaultLanguageAdapter;
-import org.quiltmc.loader.impl.util.StringUtil;
 import org.quiltmc.loader.impl.util.SystemProperties;
 import net.fabricmc.accesswidener.AccessWidener;
 import net.fabricmc.accesswidener.AccessWidenerReader;
@@ -123,13 +122,13 @@ public class QuiltLoaderImpl implements FabricLoader {
 	private void setGameDir(Path gameDir) {
 		this.gameDir = gameDir;
 		String configDir = System.getProperty(SystemProperties.CONFIG_DIRECTORY);
-		this.configDir = gameDir.resolve(StringUtil.either(configDir, DEFAULT_CONFIG_DIR));
-		refreshModsDir(gameDir);
+		this.configDir = gameDir.resolve((configDir == null || configDir.isEmpty()) ? DEFAULT_CONFIG_DIR : configDir);
+		initializeModsDir(gameDir);
 	}
 
-	private void refreshModsDir(Path gameDir) {
+	private void initializeModsDir(Path gameDir) {
 		String modsDir = System.getProperty(SystemProperties.MODS_DIRECTORY);
-		this.modsDir = gameDir.resolve(StringUtil.either(modsDir, DEFAULT_MODS_DIR));
+		this.modsDir = gameDir.resolve((modsDir == null || modsDir.isEmpty()) ? DEFAULT_MODS_DIR : modsDir);
 	}
 
 	@Override
@@ -171,7 +170,7 @@ public class QuiltLoaderImpl implements FabricLoader {
 			try {
 				Files.createDirectories(configDir);
 			} catch (IOException e) {
-				throw new RuntimeException("Failed to create config directory", e);
+				throw new RuntimeException(String.format("Failed to create config directory at '%s'", configDir), e);
 			}
 		}
 		return configDir;
@@ -184,16 +183,16 @@ public class QuiltLoaderImpl implements FabricLoader {
 	}
 
 	public Path getModsDir() {
+		// modsDir should be initialized before this method is ever called, this acts as a very special failsafe
 		if (modsDir == null) {
-			// Should not be null ever
-			refreshModsDir(gameDir);
+			initializeModsDir(gameDir);
 		}
 
 		if (!Files.exists(modsDir)) {
 			try {
 				Files.createDirectories(modsDir);
 			} catch (IOException e) {
-				throw new RuntimeException("Failed to create mods directory", e);
+				throw new RuntimeException(String.format("Failed to create mods directory at '%s'", modsDir), e);
 			}
 		}
 		return modsDir;
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -119,8 +119,8 @@ public class QuiltLoaderImpl implements FabricLoader {
 
 	private void setGameDir(Path gameDir) {
 		this.gameDir = gameDir;
-		String configSubDir = System.getProperty(SystemProperties.CONFIG_SUBDIRECTORY);
-		this.configDir = resolveIfNotAbsolute(gameDir, configSubDir == null || configSubDir.isEmpty()  ? "config" : configSubDir);
+		String configDir = System.getProperty(SystemProperties.CONFIG_DIRECTORY);
+		this.configDir = resolveIfNotAbsolute(gameDir, configDir == null || configDir.isEmpty()  ? "config" : configDir);
 		refreshModsDir(gameDir);
 	}
 
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -22,7 +22,6 @@ import java.io.IOException;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -16,51 +16,38 @@
 
 package org.quiltmc.loader.impl;
 
+import net.fabricmc.accesswidener.AccessWidener;
+import net.fabricmc.accesswidener.AccessWidenerReader;
+import net.fabricmc.api.EnvType;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.LanguageAdapter;
+import net.fabricmc.loader.api.MappingResolver;
+import net.fabricmc.loader.api.SemanticVersion;
+import net.fabricmc.loader.api.entrypoint.EntrypointContainer;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.ApiStatus;
+import org.objectweb.asm.Opcodes;
+import org.quiltmc.loader.impl.discovery.*;
+import org.quiltmc.loader.impl.game.GameProvider;
+import org.quiltmc.loader.impl.gui.QuiltGuiEntry;
+import org.quiltmc.loader.impl.launch.common.QuiltLauncher;
+import org.quiltmc.loader.impl.launch.common.QuiltLauncherBase;
+import org.quiltmc.loader.impl.launch.knot.Knot;
+import org.quiltmc.loader.impl.metadata.DependencyOverrides;
+import org.quiltmc.loader.impl.metadata.EntrypointMetadata;
+import org.quiltmc.loader.impl.metadata.LoaderModMetadata;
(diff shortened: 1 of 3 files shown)


src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -117,7 +118,14 @@ public void setGameProvider(GameProvider provider) {
 
 	private void setGameDir(Path gameDir) {
 		this.gameDir = gameDir;
-		this.configDir = gameDir.resolve("config");
+		String configSubDir = System.getProperty("quilt.configSubDir");
```
i509VCB, line 121: We have a `SystemProperties` class for a reason...

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -172,7 +180,19 @@ public File getConfigDirectory() {
...
+		if (modDir == null) {
+			// Should not be null ever
+			refreshModDir(gameDir);
+		}
+
+		if (!Files.exists(modDir)) {
+			try {
+				Files.createDirectories(modDir);
+			} catch (IOException e) {
+				throw new RuntimeException("Creating mods directory", e);
```
ByMartrixx, line 192:
```suggestion
				throw new RuntimeException("Failed to create mods directory", e);
```
I think this should be more descriptive to what the error is

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -117,7 +118,14 @@ public void setGameProvider(GameProvider provider) {
 
 	private void setGameDir(Path gameDir) {
 		this.gameDir = gameDir;
-		this.configDir = gameDir.resolve("config");
+		String configSubDir = System.getProperty("quilt.configSubDir");
+		this.configDir = gameDir.resolve(configSubDir == null ? "config" : configSubDir);
```
ByMartrixx, line 122:
```suggestion
		this.configDir = gameDir.resolve(configSubDir == null || configSubDir.isEmtpy() ? "config" : configSubDir);
```
This may be empty if left as `-Dquilt.configSubDir` with no value

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -117,7 +118,14 @@ public void setGameProvider(GameProvider provider) {
...
+		refreshModDir(gameDir);
+	}
+
+	private void refreshModDir(Path gameDir) {
+		String modSubDir = System.getProperty("quilt.modSubDir");
+		this.modDir = gameDir.resolve(modSubDir == null ? "mods" : modSubDir);
```
ByMartrixx, line 128:
```suggestion
		this.modDir = gameDir.resolve(modSubDir == null || modSubDir.isEmpty() ? "mods" : modSubDir);
```

ByMartrixx (changes requested):

AlexIIL:
Is there a reason why this mandates both folders as being *sub* paths of the game path, rather than being anywhere else? I just find that a little odd, that's all.

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
FoundationGames: This is an exact copy of the error thrown when creating the config folder fails. Both will have to be changed.

ByMartrixx (approved):

FoundationGames:
> Is there a reason why this mandates both folders as being sub paths of the game path, rather than being anywhere else? I just find that a little odd, that's all.
@AlexIIL Mainly that you'd have to additionally configure whether the specified path is to be a subfolder or not, and I didn't think that would be necessary. If you have an idea on how this could be implemented, let me know.

IMS212 (approved):

Haven-King:
> @AlexIIL Mainly that you'd have to additionally configure whether the specified path is to be a subfolder or not, and I didn't think that would be necessary. If you have an idea on how this could be implemented, let me know.

When parsing paths, Java will recognize if a path is absolute or not. As such, something like this could be done to simply allow users to provide either relative *or* absolute paths:
```java
  private void setGameDir(Path gameDir) {
    this.gameDir = gameDir;
    this.configDir = resolveDirectory(gameDir, SystemProperties.CONFIG_SUBDIRECTORY, "config");
    refreshModDir(gameDir);
  }

  private void refreshModDir(Path gameDir) {
    this.modDir = resolveDirectory(gameDir, SystemProperties.MOD_SUBDIRECTORY, "mods");
  }

  private Path resolveDirectory(Path gameDir, String systemProperty, String defaultName) {
    String subDir = System.getProperty(systemProperty);

    if (subDir == null || subDir.isEmpty()) {
      return gameDir.resolve(defaultName);
    } else {
      Path path = Paths.get(subDir);

      return path.isAbsolute() ? path : gameDir.resolve(path);
    }
  }
```

Kroppeb:
@FoundationGames @AlexIIL @Haven-King 
> ```java
> return path.isAbsolute() ? path : gameDir.resolve(path);
> ```
is the default behaviour of `resolve`. So this should already be handled.

i509VCB:
> > @AlexIIL Mainly that you'd have to additionally configure whether the specified path is to be a subfolder or not, and I didn't think that would be necessary. If you have an idea on how this could be implemented, let me know.
> 
> When parsing paths, Java will recognize if a path is absolute or not. As such, something like this could be done to simply allow users to provide either relative _or_ absolute paths:
> 
> ```java
>   private void setGameDir(Path gameDir) {
>     this.gameDir = gameDir;
>     this.configDir = resolveDirectory(gameDir, SystemProperties.CONFIG_SUBDIRECTORY, "config");
>     refreshModDir(gameDir);
>   }
> 
>   private void refreshModDir(Path gameDir) {
>     this.modDir = resolveDirectory(gameDir, SystemProperties.MOD_SUBDIRECTORY, "mods");
>   }
> 
>   private Path resolveDirectory(Path gameDir, String systemProperty, String defaultName) {
>     String subDir = System.getProperty(systemProperty);
> 
>     if (subDir == null || subDir.isEmpty()) {
>       return gameDir.resolve(defaultName);
>     } else {
>       Path path = Paths.get(subDir);
> 
>       return path.isAbsolute() ? path : gameDir.resolve(path);
>     }
>   }
> ```

Haven, I assume you are referring to impl methods there?

We can't have a `setModDir` method in public api because that causes a chicken and egg problem. Your mod is loaded from initial folder, but then changes to another folder?

FoundationGames:
> Haven, I assume you are referring to impl methods there?
> 
> We can't have a `setModDir` method in public api because that causes a chicken and egg problem. Your mod is loaded from initial folder, but then changes to another folder?

@i509VCB 

None of the methods related to setting the mod directory are visible or in the public api (At least, they shouldn't be)

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -117,7 +119,28 @@ public void setGameProvider(GameProvider provider) {
...
+	 * returned as a path. If it is not absolute, it will be returned
+	 * as a subdirectory of the passed rootDir.
+	 *
+	 * @param rootDir The parent directory to use if toResolve is not absolute
+	 * @param toResolve The subdirectory/absolute path to be interpreted
+	 * @return Either a path leading to a subdirectory of rootDir, or an absolute path
+	 */
+	private Path resolveIfNotAbsolute(Path rootDir, String toResolve) {
+		Path abs = Paths.get(toResolve);
+		return abs.isAbsolute() ? abs : rootDir.resolve(toResolve);
```
AlexIIL, line 143: From the documentation of `Path.resolve(Path other)`:

````java
/** If the {@code other} parameter is an {@link #isAbsolute() absolute}
 * path then this method trivially returns {@code other}. */
````

As such you can replace this entire method with just a call to `Path.resolve(String other)`, since that method *also* inherits the same behaviour. (And it obtains a `Path` using basically the same logic as `Paths.get(String)`)

------------

This *also* means my [earlier comment](https://github.com/QuiltMC/quilt-loader/pull/20#issuecomment-855326673) was wrong - the original code already supported absolute paths, and [this change](https://github.com/QuiltMC/quilt-loader/pull/20/commits/72be98c6e64e2c527f87e0b01a0792bed2e2d389#diff-96a52e7f80ce90440904f48ddd2fc4209f1875799268a24aff53471f8f06ef3cL122-R143) wasn't necessary. Sorry :/

(However the name change of the system property is probably a good thing - although I guess we'll have to document that the property can be used for both absolute and sub-paths)?

Comment (FoundationGames) [reply src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java:143]: I noticed this in others' comments a while after the last commit, and didn't have much luck finding documentation to back it up, now I know to revert this change. Thanks

**Review by AlexIIL — APPROVED** [review]

**Review by Haven-King — APPROVED** [review]

File: src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java [hunk 6]
```diff
@@ -117,7 +118,14 @@ public void setGameProvider(GameProvider provider) {
 
 	private void setGameDir(Path gameDir) {
 		this.gameDir = gameDir;
-		this.configDir = gameDir.resolve("config");
+		String configDir = System.getProperty(SystemProperties.CONFIG_DIRECTORY);
+		this.configDir = gameDir.resolve(configDir == null || configDir.isEmpty()  ? "config" : configDir);
```
OroArmor, line 122:
```suggestion
public static final String DEFAULT_CONFIG_PATH = "config";
private static boolean isEmptyOrNull(String string) {
    return string== null || string.isEmpty();
}
...
		this.configDir = gameDir.resolve(isEmptyOrNull(configDir)  ? DEFAULT_CONFIG_PATH  : configDir);
```

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -117,7 +118,14 @@ public void setGameProvider(GameProvider provider) {
...
+		refreshModsDir(gameDir);
+	}
+
+	private void refreshModsDir(Path gameDir) {
+		String modsSubDir = System.getProperty(SystemProperties.MODS_DIRECTORY);
+		this.modsDir = gameDir.resolve(modsSubDir == null || modsSubDir.isEmpty() ? "mods" : modsSubDir);
```
OroArmor, line 128:
```suggestion
public static final String DEFAULT_MODS_PATH = "mods";
...
		this.modsDir= gameDir.resolve(isEmptyOrNull(modsSubDir)  ? DEFAULT_MODS_PATH : modsSubDir);
```

OroArmor (changes requested):
maybe make some constants for these. also the `isNullOrEmpty` is just a suggestion. Also don't merge this directly in, it wont be valid.

OroArmor (approved):

i509VCB:
The tenary operator seemed fine, with a single use I don't see why we need a dedicated util method

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -159,7 +171,7 @@ public Path getConfigDir() {
 			try {
 				Files.createDirectories(configDir);
 			} catch (IOException e) {
-				throw new RuntimeException("Creating config directory", e);
+				throw new RuntimeException("Failed to create config directory", e);
```
i509VCB, line 174: Hopefully not too big of an interjection, could you include the stringified path in the stacktrace so we know where the mods folder creation has failed if the property is passed

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -172,7 +184,19 @@ public File getConfigDirectory() {
 	}
 
 	public Path getModsDir() {
-		return getGameDir().resolve("mods");
+		if (modsDir == null) {
+			// Should not be null ever
+			refreshModsDir(gameDir);
```
i509VCB, line 189: This could return a wrong value if someone changes the property of the mods dir via [System.setProperty()](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#setProperty(java.lang.String,%20java.lang.String)).

We should resolve the default folder, whether it is default location or via system property before handing off control to any mods or mixin config plugins.

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -117,7 +122,14 @@ public void setGameProvider(GameProvider provider) {
 
 	private void setGameDir(Path gameDir) {
```
i509VCB, line 123: This seems to imply the system property overrules a request by the game provider for the location of the game dir, typically not called in modern mc.

I'd guess you intended that the system property would overrule here.

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -117,7 +122,14 @@ public void setGameProvider(GameProvider provider) {
...
 		this.gameDir = gameDir;
-		this.configDir = gameDir.resolve("config");
+		String configDir = System.getProperty(SystemProperties.CONFIG_DIRECTORY);
+		this.configDir = gameDir.resolve(StringUtil.either(configDir, DEFAULT_CONFIG_DIR));
+		refreshModsDir(gameDir);
+	}
+
+	private void refreshModsDir(Path gameDir) {
```
i509VCB, line 130: Method name seems wrong.

src/main/java/org/quiltmc/loader/impl/util/StringUtil.java
```diff
@@ -29,4 +32,8 @@ private StringUtil() {
 			return new String[] { defaultNamespace, s };
 		}
 	}
+
+	public static String either(@Nullable String candidate, @NotNull String orElse) {
+		return (candidate == null || candidate.isEmpty()) ? orElse : candidate;
```
i509VCB, line 37: A tenary operator or an if statement should suffice here. This util method doesnt see enough use imo for it.

i509VCB (changes requested):
A few things I noticed

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
AlexIIL: AFAIK this will only be set once (since nothing sets the `modsDir` field to null) so we don't have to worry about that - after it has been called the first time (like when scanning for mods) it can't change. (Which is also what we want, since only loader plugins should be able to change the mods directory in the future, and definitely not at runtime)

i509VCB: So Alex, the value is set lazily. When someone first queries the mods directory, this is called. So the first person to set the system property and call `getModsDir` can change the return value to be incorrect. Easily done via pre launch if no one else is before you or in a mixin plugin.

AlexIIL: The `getModsDir()` method is called to determine what mods to load - I.E. in order to determine what mixin plugins might be available, or what pre-launch entrypoints need to be called. So it's set lazily, yes - however it's set *before* either of those are even loaded.

i509VCB: Okay, what about the config dir then? It seems to suffer a similar impl case but without the early loading that mod resolution does.

AlexIIL: Currently `DependencyOverrides` calls it, also before mods are fully loaded.

i509VCB: Hmm, okay so all cases are covered.

FoundationGames: > So Alex, the value is set lazily. When someone first queries the mods directory, this is called. So the first person to set the system property and call `getModsDir` can change the return value to be incorrect. Easily done via pre launch if no one else is before you or in a mixin plugin.

Nope, that's a fail safe in case something cursed happens since the mod directory is set and initialized right as the loader impl instance is setup, way before mod loading

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
FoundationGames: How's `initializeModDir()`?

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
FoundationGames: Not really sure what you mean there, none of this code touches the setting of the game directory itself whatsoever

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
i509VCB: Seems fine to me

PR #21, opened 2021-06-06
Fix provided mods showing up multiple times.

AlexIIL:
Neither the simple nor advanced code-paths actually needed provided mods (the simple path ignored them during iteration, and the advanced path handled them separately already), however they are needed for validation right at the end, so I preserved the original behaviour in a separate map (`fullCandidateMap`).

This example uses the same mod as my example from https://github.com/QuiltMC/quilt-loader/pull/11#issuecomment-855308489, however this time I didn't manually change the loader version in fabric.mod.json to `0.9.0`.

Before:
image

After:
image


TheGlitch76 (approved):

Haven-King (approved):

AlexIIL:
Since this is a simple bugfix and no-one has had any objections, I'll merge this today at 23:00 UTC.

*(I saw some comments about a proper system for who/what/when things like this should be merged, but nothing definite quite yet).*

PR #23, opened 2021-06-08
Take loom out of main buildscript

i509VCB:
This moves the minecraft testmod into a subproject for the time being and makes running the client in dev much easier.

This needs some testing to make sure the published loader artifact still works since MC implies some dependencies.

Also this adds the license header for quilt in the format we want, since we have created the subproject for testing mc. We need to determine how to approach the rest of the project later regarding licenses.


TheGlitch76 (approved):
I'm fine with testing after merging this one, considering nobody is using Loader right now.

i509VCB:
Going to test this now then.

PR #24, opened 2021-06-09
Fix provided mods.

i509VCB:
The issue here was the `ModDep` would not resolve the true "root" of the load option.

A load option is a mod that can or may be loaded. The issue here was that a provided mod load option (`ProvidedModOption`) was added to the dependency helper rather than the root, which represents the mod providing the provided mod. Because of this bug, the provided mod option was not considered as a duplicate, which meant that the provided mod and real mod were considered to be part of the compatible mod set. This means when `addMod` is called, the candidate in the load option for the real mod was added twice.

I think my description of the issue isn't random psychobabble, maybe a little roundabouty from debugging for an hour and a half.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/discovery/ModResolver.java
+++ b/src/main/java/org/quiltmc/loader/impl/discovery/ModResolver.java
@@ -1612,16 +1612,20 @@ public class ModResolver {
 
 		@Override
 		ModDep put(DependencyHelper<LoadOption, ModLink> helper) throws ContradictionException {
-			List<LoadOption> clause = new ArrayList<>(validOptions.size());
+			LoadOption[] allowed = new LoadOption[validOptions.size() + 1];
+			int i = 0;
 
-			for (ModLoadOption validOption : validOptions) {
-				// Obtain the root of the valid options to resolve mod dependencies that succeed because the depended mod is
-				// provided.
-				clause.add(validOption.getRoot());
+			for (; i < validOptions.size(); i++) {
+				// We don't want to have multiple "variables" for a single mod, since a provided mod is always present
+				// if the providing mod is present and vice versa. Calling getRoot means we depend on the provider and
+				// therefore solve properly rather than potentially saying the provided mod is present but the provide
+				// isn't and vice versa.
+				allowed[i] = validOptions.get(i).getRoot();
 			}
 
-			clause.add(new NegatedLoadOption(source));
-			helper.clause(this, clause.toArray(new LoadOption[0]));
+			// i is incremented when we exit the for loop, so this is fine.
+			allowed[i] = new NegatedLoadOption(source);
+			helper.clause(this, allowed);
 			return this;
 		}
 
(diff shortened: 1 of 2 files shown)


LambdAurora (approved):

i509VCB:
Oh yeah also one of the tests broke, see `providedDepends`.

src/main/java/org/quiltmc/loader/impl/discovery/ModResolver.java
```diff
@@ -1611,8 +1612,14 @@ public ModDep(Logger logger, ModLoadOption source, ModDependency publicDep, ModI
 
 		@Override
 		ModDep put(DependencyHelper<LoadOption, ModLink> helper) throws ContradictionException {
-			List<LoadOption> clause = new ArrayList<>();
-			clause.addAll(validOptions);
+			List<LoadOption> clause = new ArrayList<>(validOptions.size());
```
AlexIIL, line 1615: Should be:
```suggestion
			List<LoadOption> clause = new ArrayList<>(validOptions.size() + 1);
```
Since we add a negated option too. Although if you're trying to optimise then it would be better to replace `ArrayList<LoadOption>` with `LoadOption[]` and use an index. But that's really not necessary.

src/main/java/org/quiltmc/loader/impl/discovery/ModResolver.java
```diff
@@ -1712,6 +1719,8 @@ public ModBreakage(Logger logger, ModLoadOption source, ModDependency publicDep,
 		ModBreakage put(DependencyHelper<LoadOption, ModLink> helper) throws ContradictionException {
 			LoadOption[] disallowed = new LoadOption[invalidOptions.size()];
 			for (int i = 0; i < invalidOptions.size(); i++) {
+				// Obtain the root of the invalid options to resolve mod breakages that occur because a mod that breaks
+				// is provided.
 				disallowed[i] = invalidOptions.get(i).getRoot();
```
AlexIIL, line 1724: Technically the reason is that we don't want to have multiple "variables" for a single mod - as a provided mod is always present if the providing mod is present, and always missing if the providing mod is not present. Using `getRoot` ensures we actually depend on the provider (and thus we solve correctly) rather than potentially saying that a "provided" mod is present but the provider isn't (or vice versa)

AlexIIL (changes requested):
Looks functionally correct to me, only minor issues.

AlexIIL:
I think the tests are best fixed later, but temporarily commenting out [ModResolvingTests line 88](https://github.com/QuiltMC/quilt-loader/blob/master/src/test/java/net/fabricmc/test/ModResolvingTests.java#L88) is a quick-fix. (Since provided mods would no longer be returned from `ModResolver.findCompatibleSet`, instead being handled by code that the tests don't call - although i want to fix this soon after this PR).

src/main/java/org/quiltmc/loader/impl/discovery/ModResolver.java
i509VCB: Resolved with array

src/main/java/org/quiltmc/loader/impl/discovery/ModResolver.java
i509VCB: I put your description in there since it seems to be more comprehensible than mine.

i509VCB:
Alex said it looks good on github. If there are any other problems, open a new issue.

PR #25, opened 2021-06-10
Fix default adapter method entrypoints' default and object method calls

liach:
Also moves tests to dedicated quilt package

Note: without this change, the new test will fail on the addition of the first method-based entrypoint, as method-based ones have broken object methods and cannot be added to hash sets.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/util/DefaultLanguageAdapter.java
+++ b/src/main/java/org/quiltmc/loader/impl/util/DefaultLanguageAdapter.java
@@ -16,6 +16,9 @@
 
 package org.quiltmc.loader.impl.util;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandleProxies;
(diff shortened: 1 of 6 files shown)


gdude2002:
The developer that submitted this PR has been blocked from interacting with this GitHub organisation. For the sake of transparency, we thought it appropriate to state that this was **not** due to the content submitted in the PR.

The community team elected to block the submitter due to actions they've taken in other, related communities. Specific concerns include the apparent endorsement of lethal force used by governments to oppress protestors, as well as posting discriminatory material encouraging people to exclude members of the LGBT community and people with disabilities from their spaces.

Actions like this break our [Code of Conduct](https://quiltmc.org/community/code-of-conduct.html), which also provides a provision for us to take action when people break it outside our community spaces. This provision exists to allow us to protect members of minority groups in our community, by pre-empting issues before they occur.

---

Some of the development team members pointed out that we should have written this up and posted it at the time of blocking - we'll make sure we do this promptly in future.

PR #28, opened 2021-06-10
Make ModResolver.resolve return ModResolveResult

AlexIIL:
This makes it possible to validate provided mods in the tests, and makes it *very slightly* easier to add more data to the result (perhaps custom `ModLink`s or `LoadOption`s for loader plugins, or possibly explanations for *why* a given mod candidate was or was not chosen - either need lots more work to actually implement).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ out/
 .classpath
 .project
 .settings/
+*.launch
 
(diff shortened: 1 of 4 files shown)


AlexIIL:
Closing since I'll push this as part of QMJ.

PR #29, opened 2021-06-10
Draft impl of passing a ModContainer to the builtin entrypoints

### LINKED ISSUE #26: Better entrypoints
This issue might be closed once entrypoints are not in loader anymore.

Entrypoints should have a `ModContainer` parameter passed to them, I don't know why it wasn't from the start but it can be pretty annoying in the case a mod needs to get its `ModContainer` as it has to get it with a call using its own mod id which returns an `Optional`.

This mostly have a use-case for resource loader and the built-in resource packs, could be used for other things too.
It's also a way to get the root `Path` of the mod.

AlexIIL:
Draft for 2 reasons:

* There's not been any discussion on #26 yet. *(Although I like the proposal, hence this impl draft)*
* I haven't tested this on real mods with both kinds of entrypoints.

Since quilt uses a different namespace we can just add new interfaces under `org.quiltmc.loader.api`. (I didn't put these under `org.quilt.api` since I'm not sure what the eventual package structure will be for loader plugins and QSL - and it's probably better to move these *out* from loader rather than pollute a package that might be used in the future).

Discussion about the original request (I.E. adding a `ModContainer` parameter to entrypoints) should go in #26, but issues with this implementation of it should go here.

This closes #26.

### DIFF
--- a/src/main/java/net/fabricmc/api/DedicatedServerModInitializer.java
+++ b/src/main/java/net/fabricmc/api/DedicatedServerModInitializer.java
@@ -16,6 +16,8 @@
 
 package net.fabricmc.api;
 
+import net.fabricmc.loader.api.ModContainer;
+
 /**
  * A mod initializer ran only on {@link EnvType#SERVER}.
  *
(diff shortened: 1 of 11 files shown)


AlexIIL:
I think the main issue with this is that it currently points to fabric's `ModContainer` rather than quilt's - although this requires us to clean up quilt's systems (since currently we still use a lot of fabric's interfaces and don't have equivalents for quilt, and I'm not quite sure how that will end).

However that hasn't been started yet, so this can probably be

PR #30, opened 2021-06-20
Remove v0 mod metadata and the things surrounding it

TheGlitch76:
Upstream broke everything that would use v0 mod metadata by migrating the impl details of loader to a different package.
We may as well remove this technical debt now.

This PR is a precursor to some incoming Loader Plugin-style abstractions to mod loading, which are needed to make the QMJ stuff not a complete mess.


src/main/java/org/quiltmc/loader/impl/metadata/ModMetadataParser.java
```diff
@@ -116,10 +116,10 @@ private static LoaderModMetadata readModMetadata(Logger logger, JsonReader reade
 		case 1:
 			return V1ModMetadataParser.parse(logger, reader);
 		case 0:
-			return V0ModMetadataParser.parse(logger, reader);
+			throw new ParseMetadataException("Quilt Loader does not support the old-style v0 of the fabric.mod.json specification.");
 		default:
 			if (schemaVersion > 0) {
-				throw new ParseMetadataException(String.format("This version of fabric-loader doesn't support the newer schema version of \"%s\""
+				throw new ParseMetadataException(String.format("This version of Quilt Loader doesn't support the newer schema version of \"%s\""
 					+ "\nPlease update fabric-loader to be able to read this.", schemaVersion));
```
Platymemo, line 123:
```suggestion
					+ "\nPlease update Quilt Loader to be able to read this.", schemaVersion));
```

Platymemo (changes requested):

AlexIIL (commented):
Personally I'm not very happy to see backwards compatibility be removed from loader, since loader is meant to work on all versions (old and new), unlike QSL.

However this functionality would need to be re-implemented anyway in a fabric-compat plugin, and that can by done by copying the relavent code directly from fabric-loader, so I don't think removing this from here is a bad idea.

As a sidenote, this *isn't technical debt*, since none of this code is really limiting our ability to make changes. (As we need to keep FMJ v1 around, and the code for handling v1 is already most of the way there towards handling v0 as well).

Leo40Git (approved):
goodbye old code, we knew ye well :pensive:

Siuolthepic (approved):
I support removal of v0 mod metadata.

BTW, will this bump us a version? Because with some of the more recent commits it would be really nice to see this in the maven.

PR #31, opened 2021-06-21
QMJ partial impl

AlexIIL:
Misc work. I haven't finished yet, but we're a lot closer to quilt mods being able to load than before.
image
(*Notice the "quilt" icon next to the jar :)* )

I haven't even written *fallback* error descriptions yet, hence the todo's.

I'll merge this soon, although I'm not sure if we need to squash it?


AlexIIL:
(This is only a PR rather than a direct push since this steps pretty heavily on QMJ as a whole).

src/experimental/java/org/quiltmc/loader/impl/VersionConstraintImpl.java
```diff
@@ -60,4 +82,40 @@ public boolean equals(Object o) {
...
+						&& semanticVersion.minor() == semantic.minor();
+				default:
+					throw new IllegalStateException("Unknown VersionConstraint.Type " + type);
+			}
+		} else {
+			if (version.raw().equals("${version}")) {
+				// Special cased by QMJ
+				return true;
+			}
+			return false;
```
TheGlitch76, line 118: This returns false for a.matches(a)

src/experimental/java/org/quiltmc/loader/api/ConvertibleModMetadata.java
```diff
@@ -0,0 +1,11 @@
+package org.quiltmc.loader.api;
+
+/** @deprecated Since this is very much not a good API, and needs changing.
+ *             <p>
+ *             Or at least re-evaluted later on to see if this should be kept as-is. */
+@Deprecated
+public interface ConvertibleModMetadata {
+    net.fabricmc.loader.api.metadata.ModMetadata asFabricModMetadata();
```
TheGlitch76, line 8: this *is* gross, but it's fine because abstracting all this away will be its own project

i509VCB: This shouldn't be api imo.

src/experimental/java/org/quiltmc/loader/impl/VersionConstraintImpl.java
```diff
@@ -60,4 +82,40 @@ public boolean equals(Object o) {
...
+				case SAME_MAJOR:
+					return semanticVersion.major() == semantic.major();
+				case SAME_MAJOR_AND_MINOR:
+					return semanticVersion.major() == semantic.major()
+						&& semanticVersion.minor() == semantic.minor();
+				default:
+					throw new IllegalStateException("Unknown VersionConstraint.Type " + type);
+			}
+		} else {
+			if (version.raw().equals("${version}")) {
```
i509VCB, line 114: This should be dev only really

src/experimental/java/org/quiltmc/loader/api/ConvertibleModMetadata.java
AlexIIL: Actually it would probably be cleaner if this was moved outside of api too :p

src/experimental/java/org/quiltmc/loader/impl/VersionConstraintImpl.java
AlexIIL: Yes, that's a good point.

src/experimental/java/org/quiltmc/loader/impl/VersionConstraintImpl.java
AlexIIL: How?

`VersionConstraintImpl` doesn't implement `Version`, so it's not possible to pass the constaint back into itself (`VersionConstraintImpl.matches(Version)`).

(Although I do think it's a bit off that we can't handle string versions here, however that's what the QMJ spec suggests?)

TheGlitch76 (approved):
Good enough for me. This is all experimental API, let's get a prototype in.

AlexIIL:
Should we wait for a second review? (If we started following https://github.com/QuiltMC/rfcs/pull/25 now at least, we'd need a second one).

Either way I'm not going to make any further changes for now.

TheGlitch76:
i5 gave me permission to push whatever into the qmj branch so I imagine it's fine. Might be worth pinging him on discord?

AlexIIL:
From @i509VCB in #loader-general:
> [21:11] i509VCB: I'd say it's fine
[21:11] i509VCB: I am going to do a big testing operation before we unveil it for use

So I'll merge this now.

PR #32, opened 2021-06-21
Proof of Concept: Move Minecraft-specific code to its own project

TheGlitch76:
I don't intend this to be merged in its current state, but I want to show our current progress. As such, this description will be updated often.

This branch will serve as the base for Loader Plugins. Currently it doesn't do anything resembling an actual plugin, it is only a group of two ServiceLoaders that load the `GameProvider` and `Transformer` classes.

I removed some legacy code involving LaunchWrapper, and some code involving compatibility with (roughly) 1.9 to 1.12 FML. For the latter, I added an API to GameProvider that would allow it to be re-added by a plugin.


TheGlitch76:
handled by upstream in upcoming merge

PR #36, opened 2021-07-01
Quilt.mod.json prototype

TheGlitch76:
I'm just opening this so we can have one place to put comments. This is nowhere near done.


src/experimental/java/org/quiltmc/loader/impl/metadata/qmj/ModLicenseImpl.java
```diff
@@ -0,0 +1,85 @@
...
+		}
+	}
+
+	public static @Nullable ModLicense fromIdentifier(String identifier) {
+		return licenses.get(identifier);
+	}
+
+	public static ModLicense fromIdentifierOrDefault(String identifier) {
+		ModLicense ret = licenses.get(identifier);
+		if (ret == null) {
```
i509VCB, line 52: I like spaced things

src/experimental/java/org/quiltmc/loader/impl/metadata/qmj/ModLicenseImpl.java
```diff
@@ -0,0 +1,85 @@
...
+			return new ModLicenseImpl(identifier, identifier, "", "");
+		} else {
```
i509VCB, line 54: unnesscary else block

src/main/java/org/quiltmc/loader/impl/discovery/ClasspathModCandidateFinder.java
```diff
@@ -46,15 +46,10 @@ public void findCandidates(QuiltLoaderImpl loader, BiConsumer<URL, Boolean> appe
...
-				while (mods.hasMoreElements()) {
-					try {
-						modsList.add(UrlUtil.getSource("fabric.mod.json", mods.nextElement()));
-					} catch (UrlConversionException e) {
-						loader.getLogger().debug(e);
-					}
-				}
+
+				addModSources(loader, modsList, "quilt.mod.json");
+				addModSources(loader, modsList, "fabric.mod.json");
```
i509VCB, line 52: could this result in double adding of mods which contain both files?

src/main/java/org/quiltmc/loader/impl/discovery/ClasspathModCandidateFinder.java
```diff
@@ -116,4 +111,15 @@ public void findCandidates(QuiltLoaderImpl loader, BiConsumer<URL, Boolean> appe
 			}
 		});
 	}
+
+	protected void addModSources(QuiltLoaderImpl loader, Set<URL> modsList, String name) throws IOException {
+		Enumeration<URL> mods = QuiltLauncherBase.getLauncher().getTargetClassLoader().getResources(name);
+		while (mods.hasMoreElements()) {
```
i509VCB, line 117: I like spaced things. (Put a space between the field declaration and while block)

src/main/java/org/quiltmc/loader/impl/discovery/ClasspathModCandidateFinder.java
AlexIIL: I don't think so, since it has a `Set<URL>`, so it will call `URL.hashCode()` and `URL.equals()` to check that they are same. (And the URL's are of the *base* resource, not the `*.mod.json` file so comparing them will be useful).

Wait a minute, isn't `URL.equals` meant to be avoided at all costs? If so then this needs re-writing anyway to avoid that.

src/test/java/org/quiltmc/test/VersionParsingTests.java
```diff
@@ -0,0 +1,92 @@
...
+		fail.beginArray();
+		while (fail.hasNext()) {
+			fabricFails(fail.nextString());
+		}
+		fail.close();
+	}
+
+
+	JsonReader get(String s) throws IOException {
+		File f = new File(getClass().getClassLoader().getResource(s).getFile());
```
Siuolthepic, line 89: The file name breaks when a parent folder has a space in its name. 

I was able to find a fix to making it `		File f = new File(getClass().getClassLoader().getResource(s).getFile().replace("%20", " "));`

Siuolthepic (approved):
Since we (according to my understanding, this is not actual fact.) are finishing up at this point, and given the fact that this code seems to work smoothly in my experience, I approve of these changes. 

My only suggestion is to bump the version to 0.14.0 after this is merged.

(Also, almost forgot to say, but fixing the License issue as well will be needed,)

AlexIIL:
I think we also need https://github.com/QuiltMC/rfcs/pull/38 to be either accepted or denied before this PR can be merged - since the qmj branch currently implements it.

PR #39, opened 2021-07-09
More QMJ work

AlexIIL:
Many changes:
 - Heavily abstracted ModLink/Rule away from sat4j directly. This is primarily useful for plugins. This *also* means that the changes I've made to `DependencyHelper` in our sat4j branch are no longer necessary.
 - Stopped using DependencyHelper since it turns out it's just not quite powerful enough for us, instead `Sat4jWrapper` has the same abstraction.
 - ~~Accidentality implemented https://github.com/QuiltMC/rfcs/pull/30 early. I'll revert that part later, or we could leave it in if no-one objects to that RFC.~~ Implemented https://github.com/QuiltMC/rfcs/pull/30
 - Removed non-advanced mode from ModSolver, since plugins will almost certainly not be compatible with it.
 - ~~Changed `ModCandidateSet.candidates` to a `ConcurrentHashMap` since it's modified by the mod-scanning threads rather than just the main thread. (To be honest I should probably have made `ModCandidateSet.add` synchronised instead)~~ Made `ModCandidateSet.add` synchronised since it's called by multiple mod-scanning threads at once, rather than just the main thread
 - Simplified mod definitions a bit.

It turns out that separating mod candidate additions and mod dependency rule defining (and instead letting rules "listen" to new mods being added) makes the code a bit simpler too, since we no longer need to pass over the mod list twice.

*(sidenote: I don't really like pushing large changes directly to the QMJ branch, but since this is targeting the QMJ branch this doesn't need to go through FCP and instead I'll just merge it soon).*


src/experimental/java/org/quiltmc/loader/impl/metadata/qmj/FabricModMetadataWrapper.java
```diff
@@ -115,6 +115,11 @@ public String description() {
 		return breaks;
 	}
 
+	@Override
+	public ModLoadType loadType() {
+		return ModLoadType.IF_POSSIBLE;
```
TheGlitch76, line 120: shouldn't this be ALWAYS? that is the behavior normally expected in fabric

AlexIIL: Nope, fabric's old mod loading behaviour was definitely `IF_POSSIBLE` for jar-in-jars, it's just that it works the same way as `ALWAYS` most of the time so people assumed that was how it works.

For example if you just jar-in-jar a library, but then conflict with it, it won't get loaded - however if it defaulted to `ALWAYS` then it would crash.

TheGlitch76: But this is a FabricModMetadataWrapper, *not* a JIJ wrapper? Unless the name for this class is awful/

AlexIIL: Oh right people don't tend to like my names :p

`FabricModMetadataWrapper` is a wrapper which converts a fabric metadata object into a quilt metadata object. So, this is effectively:

```java
public final class FabricModMetadataWrapper implements QuiltModMetadata {
  private final FabricModMetadata fabricMeta;
}
```

AlexIIL:
Sorry for the delay - currently this implements everything except provides and mod breaks. I removed the verification from ModSolver since we shouldn't actually need that anymore - since the sat4j handling should find any and all problems like those now.(With the possible exception of suspiciously similar versions).

AlexIIL:
Okay, this is probably stable enough to base other work on.

PR #44, opened 2021-09-29
Fix POM dependencies population

LambdAurora:
and move source/target java version declaration.


Siuolthepic (approved):
As little as its worth, I approve.

TheGlitch76:
Has this been tested to not break with Loom?

LambdAurora:
> Has this been tested to not break with Loom?

This PR was tested with local publishing
all artifacts are looking like they should 
and it works.

PR #48, opened 2021-11-03
Deprecate fabric's metadata, and add quilt's equivilent metadata.

AlexIIL:
Changes:

* Everything in the `net.fabricmc.api` and `net.fabricmc.loader.api` packages has been deprecated, except for `@Environment`, `@EnvironmentInterface`, `@EnvironmentInterfaces`, and `EnvType`. All usages of these classes and interfaces should get correctly redirected or used by quilt's classes. (in particular fabric mods that implement fabric's entrypoints will continue to be called by quilt's entrypoint invocations).
* `QuiltLoader` (replacing `FabricLoader`) is a static class, with only static functions.
* `getEnvironmentType()` has been moved to a minecraft specific class `MinecraftQuiltLoader`
* Entrypoints:
    * All quilt entrypoints extend a new interface `GameEntrypoint`, which is empty.
    * All entrypoint methods include the `ModContainer` as an argument (#26)
    * `ModInitializer`, `ClientModInitializer` and `DedicatedServerModInitializer` have been moved to a minecraft-specific api package (`org.quiltmc.loader.api.minecraft`)

### DIFF
--- a/src/main/java/net/fabricmc/loader/impl/quiltmc/Quilt2FabricMappingResolver.java
+++ b/src/main/java/net/fabricmc/loader/impl/quiltmc/Quilt2FabricMappingResolver.java
@@ -1,0 +1,43 @@
+package net.fabricmc.loader.impl.quiltmc;
+
+import java.util.Collection;
+
+import org.quiltmc.loader.api.MappingResolver;
+
+public class Quilt2FabricMappingResolver implements net.fabricmc.loader.api.MappingResolver {
+	private final MappingResolver quilt;
+
+	public Quilt2FabricMappingResolver(MappingResolver quilt) {
+		this.quilt = quilt;
+	}
+
+	@Override
+	public Collection<String> getNamespaces() {
+		return quilt.getNamespaces();
+	}
+
+	@Override
+	public String getCurrentRuntimeNamespace() {
+		return quilt.getCurrentRuntimeNamespace();
+	}
+
+	@Override
+	public String mapClassName(String namespace, String className) {
+		return quilt.mapClassName(namespace, className);
+	}
+
+	@Override
+	public String unmapClassName(String targetNamespace, String className) {
+		return quilt.unmapClassName(targetNamespace, className);
+	}
+
+	@Override
+	public String mapFieldName(String namespace, String owner, String name, String descriptor) {
+		return quilt.mapFieldName(namespace, owner, name, descriptor);
+	}
+
+	@Override
+	public String mapMethodName(String namespace, String owner, String name, String descriptor) {
+		return quilt.mapMethodName(namespace, owner, name, descriptor);
+	}
+}
--- a/src/main/java/net/fabricmc/loader/impl/quiltmc/Quilt2FabricMappingResolver.java
+++ b/src/main/java/net/fabricmc/loader/impl/quiltmc/Quilt2FabricMappingResolver.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2016 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package net.fabricmc.loader.impl.quiltmc;
 
 import java.util.Collection;
--- a/src/main/java/org/quiltmc/loader/api/ModInitializer.java
+++ b/src/main/java/org/quiltmc/loader/api/ModInitializer.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2016 FabricMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.quiltmc.loader.api;
 
 /**
--- a/src/main/java/org/quiltmc/loader/api/ModInitializer.java
+++ b/src/main/java/org/quiltmc/loader/api/ModInitializer.java
@@ -1,0 +1,15 @@
+package org.quiltmc.loader.api;
+
+/**
+ * A mod initializer.
+ *
+ * <p>In {@code quilt.mod.json}, the entrypoint is defined with {@code main} key.</p>
+ *
+ * @see ClientModInitializer
+ * @see DedicatedServerModInitializer
+ * @see org.quiltmc.loader.api.QuiltLoader#getEntrypointContainers(String, Class)
+ */
+@FunctionalInterface
+public interface ModInitializer {
+	void onInitialize(ModContainer mod);
+}
--- a/src/main/java/org/quiltmc/loader/api/QuiltLoader.java
+++ b/src/main/java/org/quiltmc/loader/api/QuiltLoader.java
@@ -1,34 +1,18 @@
-/*
- * Copyright 2016 FabricMC
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+package org.quiltmc.loader.api;
 
-package net.fabricmc.loader.api;
-
-import java.io.File;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 
-import net.fabricmc.api.EnvType;
-import net.fabricmc.loader.api.entrypoint.EntrypointContainer;
-import net.fabricmc.loader.impl.quiltmc.Quilt2FabricLoader;
-
 import org.jetbrains.annotations.Nullable;
-import org.quiltmc.loader.api.QuiltLoader;
+import org.quiltmc.loader.api.entrypoint.EntrypointContainer;
 import org.quiltmc.loader.impl.QuiltLoaderImpl;
+
+import net.fabricmc.loader.api.EntrypointException;
+import net.fabricmc.loader.api.LanguageAdapter;
+
+import net.fabricmc.api.EnvType;
 
 /**
  * The public-facing FabricLoader instance.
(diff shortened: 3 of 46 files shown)


src/main/java/net/fabricmc/loader/impl/quiltmc/Quilt2FabricMappingResolver.java
```diff
@@ -0,0 +1,43 @@
+package net.fabricmc.loader.impl.quiltmc;
+
+import java.util.Collection;
+
+import org.quiltmc.loader.api.MappingResolver;
+
+public class Quilt2FabricMappingResolver implements net.fabricmc.loader.api.MappingResolver {
```
OroArmor, line 23: This class will need some work once Hashed is used in loader

src/main/java/org/quiltmc/loader/api/minecraft/ClientModInitializer.java
```diff
@@ -0,0 +1,25 @@
...
+ * or integrated server tweaks.</p>
+ *
+ * <p>In {@code fabric.mod.json}, the entrypoint is defined with {@code client} key.</p>
+ *
+ * @see ModInitializer
+ * @see DedicatedServerModInitializer
+ * @see org.quiltmc.loader.api.QuiltLoader#getEntrypointContainers(String, Class)
+ */
+@FunctionalInterface
+public interface ClientModInitializer {
```
OroArmor, line 36: `extends GameEntrypoint`?

src/main/java/org/quiltmc/loader/api/minecraft/DedicatedServerModInitializer.java
```diff
@@ -0,0 +1,20 @@
...
+ * A mod initializer ran only on {@link EnvType#SERVER}.
+ *
+ * <p>In {@code fabric.mod.json}, the entrypoint is defined with {@code server} key.</p>
+ *
+ * @see ModInitializer
+ * @see ClientModInitializer
+ * @see org.quiltmc.loader.api.QuiltLoader#getEntrypointContainers(String, Class)
+ */
+@FunctionalInterface
+public interface DedicatedServerModInitializer {
```
OroArmor, line 31: same as above

src/main/java/org/quiltmc/loader/api/minecraft/ClientModInitializer.java
AlexIIL: woops, thanks!

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * <p>To obtain a working instance, call {@link #getInstance()}.</p>
+ *
+ * @since 0.4.0
+ */
+public interface QuiltLoader {
+	/**
+	 * Returns the public-facing Quilt Loader instance.
+	 */
+	static QuiltLoader getInstance() {
```
i509VCB, line 44: I would like to eliminate the `getInstance` function and make every other function accessible through quilt loader's singleton as a static function.

If a need to have a swappable impl is needed in the future, we can slap a semi-private api behind the static method calls

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
```diff
@@ -0,0 +1,191 @@
...
+import net.fabricmc.api.EnvType;
+
+/**
+ * The public-facing FabricLoader instance.
+ *
+ * <p>To obtain a working instance, call {@link #getInstance()}.</p>
+ *
+ * @since 0.4.0
+ */
+public interface QuiltLoader {
```
i509VCB, line 40: Let's think ahead of time here on plugins:

Do we need to restrict any of the functions here and they use during the plugin loading stage? If yes then the quilt loader instance should be passed via the game entrypoint (and pre launch) and the singleton kept as impl detail.

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
```diff
@@ -0,0 +1,191 @@
...
+	 * current environment type.
+	 *
+	 * <p>The game instance may not always be available depending on the game version and {@link EnvType environment}.
+	 *
+	 * @return A client or server instance object
+	 * @deprecated This method is experimental and it's use is discouraged.
+	 */
+	@Nullable
+	@Deprecated
+	Object getGameInstance();
```
i509VCB, line 169: Since not every game can fulfill this and Minecraft has changed recently to launch it's server statically, there isn't truly a `game instance` any more on both sides I think we should remove this API.

src/main/java/org/quiltmc/loader/api/ModInitializer.java
```diff
@@ -0,0 +1,31 @@
...
+ * A mod initializer.
+ *
+ * <p>In {@code quilt.mod.json}, the entrypoint is defined with {@code main} key.</p>
+ *
+ * @see ClientModInitializer
+ * @see DedicatedServerModInitializer
+ * @see org.quiltmc.loader.api.QuiltLoader#getEntrypointContainers(String, Class)
+ */
+@FunctionalInterface
+public interface ModInitializer {
```
i509VCB, line 29: These initializers (except for pre launch) should be moved to a minecraft package because not every game or application has a client, server and common mode.

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+import org.jetbrains.annotations.Nullable;
+import org.quiltmc.loader.api.entrypoint.EntrypointContainer;
+import org.quiltmc.loader.impl.QuiltLoaderImpl;
+
+import net.fabricmc.loader.api.EntrypointException;
+import net.fabricmc.loader.api.LanguageAdapter;
```
i509VCB, line 29: Incorrect imports?

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
```diff
@@ -0,0 +1,191 @@
...
+	 */
+	MappingResolver getMappingResolver();
+
+	/**
+	 * Gets the container for a given mod.
+	 *
+	 * @param id the ID of the mod
+	 * @return the mod container, if present
+	 */
+	Optional<ModContainer> getModContainer(String id);
```
i509VCB, line 121: Do we want to keep this as optional or change to nullable?

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
```diff
@@ -0,0 +1,191 @@
...
+	 * environment
+	 */
+	boolean isDevelopmentEnvironment();
+
+	/**
+	 * Get the current environment type.
+	 *
+	 * @return the current environment type
+	 */
+	EnvType getEnvironmentType();
```
i509VCB, line 155: We do need to think about game specific environment details and how to represent those without the loader having a hard requirement on the Minecraft game implementations.

src/main/java/org/quiltmc/loader/api/entrypoint/GameEntrypoint.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.quiltmc.loader.api.entrypoint;
+
+/** Marker interface for quilt entrypoints. Generally only entrypoint interfaces should extend this - and then only for
+ * entrypoints that might want to migrate to a different sort later on! */
+public interface GameEntrypoint {
```
i509VCB, line 21: Mark this as `NonExtendable`, game launch systems extending this interface is fine, but mods should not.

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
AlexIIL: Woops! Turns out those are only used for javadocs - I'll switch them out with the correct ones.

src/main/java/org/quiltmc/loader/api/ModInitializer.java
AlexIIL: Slight correction: I've already placed client&server entrypoints in the `org.quiltmc.api.minecraft` package - is that package name ok?

My assumption is that *every* game which runs quilt mods would have some stage to perform initialisation of all mods - even if it's location is different for every game. As such a single `common` initialiser would make sense for cross-game mods, even if these are fairly uncommon. (But that's technically a whole different question - can you even make cross-game mods?)

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
AlexIIL: While that's *technically* true (before the game has loaded it doesn't make sense for plugins to call certain methods) I think it would make it much worse from a usage perspective - you can only load one set of mods for minecraft, not two - so it will always make sense for `QuiltLoaderImpl` to only have one instance, and it's a lot easier to program mods if you can reliably get that instance in a static method. *(The alternative is to make all mods provide a `PreLaunchEntrypoint` just to get the QuiltLoader instance as early as possible, which sounds a bit nasty)*

Essentially I'd rather just redefine some of the methods to throw an exception if they are called too early - since there will be a lot more mods developed on quilt than plugins for quilt.

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
AlexIIL: This should be fixed now as of https://github.com/QuiltMC/quilt-loader/pull/48/commits/89f9e5cde5d6e9a785e048a5b08a139bc496ee05

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
OroArmor: `@Nullable` feels lazy, but `Optional` adds extra verbosity and time. I think the choice should be more decided on QuiltLoader acting as more functional or imperative programming

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
OroArmor: What if EnvType was some form of interface? Not sure exactly how that would work, but it would allow other games to specify their own environments

AlexIIL: Sadly `EnvType` is used as an annotation value for stripping - so we'd need to rethink how that is applied if we want to change it.

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
AlexIIL: I don't think that's a good enough reason to remove this method - since it's still useful *some of the time* just not *all* of the time. Plus, this is already documented in it's javadoc, and it's marked as `Deprecated`.

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
AlexIIL: Just to check: do you mean you want to move from:

```java
public interface QuiltLoader {
    boolean isModLoaded(String id);
}
```
to
```java
public final class QuiltLoader {
    public static boolean isModLoaded(String id) { return QuiltLoaderImpl.isModLoaded(id); }
}
```
?

Technically we can already make QuiltLoaderImpl swappable by making `QuiltLoader.getInstance` return a delegate impl of QuiltLoader and then changing what that delegate object points to, so we don't actually need to change the public api at all yet.

i509VCB: Yeah that seems fine

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
i509VCB: We could introduce a MinecraftQuiltLoader class which exposes these game specific attributes as an option?

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
AlexIIL: Ok, I quite like that idea - since it makes it much easier for mods to use directly.

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
AlexIIL: That makes more sense to me than trying to generalise this between games. Technically that could be part of a "minecraft loader plugin", but we can at least make the `MinecraftQuiltLoader` class now?

i509VCB: Yeah that seems fine for now

src/main/java/org/quiltmc/loader/api/ModInitializer.java
AlexIIL: I guess we can look into that possibility later. For now I'll just move `ModInitialiser` to the minecraft package.

AlexIIL: Moved in https://github.com/QuiltMC/quilt-loader/pull/48/commits/fa0aa003da99ea7fbbf2149748126e9e4e9204d2

src/main/java/org/quiltmc/loader/api/QuiltLoader.java
AlexIIL: Ok, implemented.

AlexIIL:
Unless anyone has any additional problems with this as-is then I'll merge this in 48 hrs, and start on loader plugins.

PR #50, opened 2021-12-02
Refactor the ModDependency model.

TheGlitch76:
These changes are in line with what upstream Fabric did to their ModDependency. However, I did change a few things:
- The method to get all mod "dependencies" is called `relations()`, because in the future it may be expanded to hold a `ModDependency` that isn't necessarily a hard dependency. (For example, we could add a `conflicts` option like fabric v1, which gives a warning screen on first launch but allows you to continue). I'd be willing to rename `relations()` to `dependencies()`, or rename `ModDependency` to `ModRelation`, if that would be clearer.
- I've removed `depends()` and `breaks()` in line with upstream, which recommends getting all relations then filtering on the `Kind` you want. I'm not sure if this is the best approach though and I'm willing to add those methods back.

I will rebase this PR onto my massive upstream update branch if approved.

### DIFF
--- a/src/main/java/org/quiltmc/loader/api/ModDependency.java
+++ b/src/main/java/org/quiltmc/loader/api/ModDependency.java
@@ -17,6 +17,7 @@
 package org.quiltmc.loader.api;
 
 import java.util.Collection;
+import java.util.Collections;
 
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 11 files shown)


AlexIIL:
I don't really understand the benefit to this? I'm not sure when we'd want to iterate over all relations and then switch on their kind rather than just iterate over each one separately?

Also, while I can understand name changes why have you removed `ModDependency.All`? If you want to consolidate then that's fine (since we'll always know what it should be based on context) but it should probably be called something other than `Any` or `All`?

TheGlitch76:
> I don't really understand the benefit to this? I'm not sure when we'd want to iterate over all relations and then switch on their kind rather than just iterate over each one separately?

tl;dr: Player, man, idk

I'm not sure why upstream decided to do this--my best guess is that upstream has so many dependency types that it tends to be less code. I'll add the `depends()` and `breaks()` methods back though; there's no good reason to remove them entirely. I want to keep the `Kind` enum and methods though, because they might come in handy when integrating with upstream's changes.

> Also, while I can understand name changes why have you removed ModDependency.All? If you want to consolidate then that's fine (since we'll always know what it should be based on context) but it should probably be called something other than Any or All?

It's more of a redefinition of terms, to be honest. This is how I define `depends` and `breaks` dependencies:
Depends: A ModDependency that must be satisfied for a mod to load.
Breaks: A ModDependency that must not be satisfied for a mod to load.
(Not in documentation, will update) Satisfied: A ModDependency with at least one `Entry` (a mod id + set of version constraints) present in the mod set

With this definition, it doesn't make sense to have a different interface type for `All`, because that would imply that `"breaks": [ "modOne", "modTwo" ]` only prevents the mod from loading if modOne *AND* modTwo are present.

TheGlitch76:
Closing this for now. Will re-evaluate once i am done merging upstream

PR #54, opened 2022-02-25
Fix Manifest Entrypoint

halotroop2288:
No such package: `org.quiltmc.impl`. It's actually `org.quiltmc.loader.impl`.


AlexIIL:
Thanks for spotting this - I guess we missed that during the package rename?

Issue #15, opened 2021-05-06
Allow Minecraft-version-specific subfolders in the mods folder


Earthcomputer:
The main use case is for people who use the vanilla launcher, which only has one mods folder shared between potentially many instances. Forge allows you to create subfolders in the mods folder named the same as each Minecraft version, and FML will not attempt to load mods from subfolders not matching your current version. I know quite a few people find it annoying that Fabric loader does not allow you to do the same thing, so it would be nice if Quilt loader had this feature.

TwoLeggedCat:
Can't you use multiple different installations with separate game directories for different versions? We wouldn't want to encourage people to use the same game directory across several versions anyway.

williambl:
IMO FML's doing this is a relic from before it was easy to separate instances into their own directories.

Perhaps we should instead make quilt installer optionally create a new directory for the quilt install?

Geolykt:
> Can't you use multiple different installations with separate game directories for different versions? We wouldn't want to encourage people to use the same game directory across several versions anyway.

It's complicated and I recall mods not working nicely with that; though that experience must be a few years old by now.

Siuolthepic:
> > Can't you use multiple different installations with separate game directories for different versions? We wouldn't want to encourage people to use the same game directory across several versions anyway.
> 
> It's complicated and I recall mods not working nicely with that; though that experience must be a few years old by now.

I have had no complications with such a thing. On Vanilla launcher its as simple as changing the installation folder to somewhere else and then dragging the mods in there instead.

chexo3:
I see no reason not to. It's simple to add and makes things easier for less technical users. We seem to forget not everyone uses MultiMC (even though they arguably should)

TwoLeggedCat:
> We seem to forget not everyone uses MultiMC (even though they arguably should)

What does this have to do with MultiMC? This is possible even with Mojang's launcher. Doing this in Quilt would just encourage people to use the same game directory across multiple versions *and* with different mods, often leadding to corruption worlds, and offer no real benefit over what the launcher itself can do.

hYdos:
>  Additionally, I'm not sure you have to be a "technical user" to put different versions in different folders, and no more so than putting different mods in the correct folders.

have you seen fabricoord's #player-support ?

TwoLeggedCat:
> > Additionally, I'm not sure you have to be a "technical user" to put different versions in different folders, and no more so than putting different mods in the correct folders.
> 
[...]

Issue #38, opened 2021-07-07
Load mods from subfolders


Reoseah:
This will allow modpack makers and users to organize their mods however they want, e.g. put optimization mods into their own subfolder like `mods/optimizations/...`

This is different from https://github.com/QuiltMC/quilt-loader/issues/15 as it just removes current limitation when searching for mods.

Currently Fabric/Quilt loader uses `1` for max depth, I suggest change it to `Integer.MAX_VALUE` in this line: https://github.com/QuiltMC/quilt-loader/blob/232e1741a3d65e4e384958cb2e18ea37b099eaf2/src/main/java/org/quiltmc/loader/impl/discovery/DirectoryModCandidateFinder.java#L58

romangraef:
Maybe this could be coupled with the ability to exclude entire folders, either based on the hidden attribute, or based on a prefix dot, like for regular mod jars as well.

Reoseah:
> Maybe this could be coupled with the ability to exclude entire folders, either based on the hidden attribute, or based on a prefix dot, like for regular mod jars as well.

Yeah, that makes sense, and hidden folders and folders starting with `.` can be some unknown shit that perhaps shouldn't be loaded. 

That could be achieved by overriding another method in the file visitor:
```java
@Override
public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
	if (dir.getFileName().toString().startsWith(".") || Files.isHidden(dir)) {
		return FileVisitResult.SKIP_SUBTREE;
	}
	return FileVisitResult.CONTINUE;
}
```

WanderingCat27:
if the loader is designed similarly to fabric in grabbing file paths as candidates it might be difficult but it would be a nice feature to search zip files as well as folders makes modpacks simple and for solves, non technical users, the mistake I see a lot of sticking a zip in the mods folder.

WanderingCat27:
> > Maybe this could be coupled with the ability to exclude entire folders, either based on the hidden attribute, or based on a prefix dot, like for regular mod jars as well.
> 
> 
> 
> Yeah, that makes sense, and hidden folders and folders starting with `.` can be some unknown shit that perhaps shouldn't be loaded. 
> 
> 
> 
> That could be achieved by overriding another method in the file visitor:
> 
> ```java
> 
> @Override
> 
> public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
> 
[...]

PR #55, opened 2022-02-27
Add a game version mod candidate finder

sschr15:
This is the simplest solution to https://github.com/QuiltMC/quilt-loader/issues/15, but we can always *make it more complex!*

...or something else if others feel like it


AlexIIL:
Hey, thanks for making this! However at the moment I'd rather gauge interest in this feature, rather than actually implement it directly - in particular this implementation works at the moment, but might not if/when subfolders are allowed - so more complex logic might need to be added if/when #38 is wanted.

*(As a sidenote: I've had a go at implementing this in the [`loader-plugins-2`](https://github.com/QuiltMC/quilt-loader/blob/loader-plugins-2/src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java#L692) branch, but this feature probably wouldn't wait until plugins to actually be implemented)*

AlexIIL:
Implemented in 0.18.1-beta.1 with loader-plugins

PR #60, opened 2022-03-31
Add `intermediate_mappings` field to `quilt.mod.json`

### LINKED ISSUE #59: Declare intermediate mappings in `quilt.mod.json`
This is a good first issue for anyone looking to dip in to contributing to Loader. If you are interested in getting this done, feel free to ask me questions at `glitch#3274` on the Toolchain Discord.

# The issue
Currently, Loader does not have a reliable way of telling what intermediate mappings were used when compiling a mod. This is a problem, because mods currently use the (legacy) Intermediary mappings, but in the future they will be compiled to Hashed.

## Motivation
We want to support Quilt mods compiled to Intermediary, even after we switch to Hashed, for two reasons:
- (major) We aren't sure when our toolchain will support compiling to Hashed, and we'd like to avoid unnecessarily breaking already compiled Quilt mods, especially since the infrastructure for remapping mods from intermediary to hashed will be present due to Fabric compatibility,
- (minor) Since (as mentioned above) we will be able to remap int -> hashed anyway, it would be nice to allow multi-loader jars (i.e. Fabric and Quilt) to use the same mapping set, to reduce file size and complexity in the build process.
    - *side note: Multi-loader jars are important to Quilt during the beta period, because it allows major mods to experiment with the new Quilt APIs without forcing their users onto an unknown, experimental, and unstable modloader.*
 
 # Proposed Solution
 This solution is intended to be as simple as possible, 
 ## QMJ Specification in RFC Repository
 Add a new, optional field to the `quilt_loader` field of the `quilt.mod.json`, named `intermediate_mappings`. It is a string value that defaults to `"hashed"`, but can be set to `"intermediary"`. 

 ## Implementation in Loader
### Parser
For now, we can simply implement a no-op in Loader that asserts that the `intermediate_mappings` field is set to `intermediary`. It should be read during mod parsing, and mod parsing should fail if:
- The value is unknown (something like "Error: Unknown intermediate mappings 'example'")
- The value is `hashed`, since we don't support it in Loader yet. This one should have a more user-friendly error message, directing the user to update Quilt Loader, since if we see this field we know that this is a mod from the future!
### Mod Metadata
The `InternalModMetadata` metadata interface should gain a new method,:
` String intermediateMappings()`
that would return the string read in the `quilt.mod.json`. For now, this can default to always return `intermediary` (see future steps below)

# Future Steps
- We will need to properly track the string given in the `quilt.mod.json` for `intermediate_mappings`,  instead of asserting that it exists and throwing the value away, so that it can be properly exposed to Loader. Bonus points if you implement it this way in your PR, but it's not required.
- In the future, we may want to allow loader plugins to support additional `intermediate_mappings` kinds, but that isn't relevant right now.

TibiNonEst:
Fixes #59

Adds the ability to declare the intermediate mappings a mod uses in the `quilt.mod.json` file.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
+++ b/src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
@@ -241,8 +241,8 @@ final class V1ModMetadataReader {
 			@Nullable
 			JsonLoaderValue intermediateMappingsValue = quiltLoader.get("intermediate_mappings");
 
-			String[] supported_mappings = { "hashed", "intermediary" };
-			String mappings;
+			String[] supported_mappings = { "org.quiltmc:hashed", "net.fabricmc:intermediary" };
+			String mappings = "org.quiltmc:hashed";
 
 			if (intermediateMappingsValue != null) {
 				if (intermediateMappingsValue.type() != LoaderValue.LType.STRING) {
@@ -251,15 +251,17 @@ final class V1ModMetadataReader {
 
 				mappings = intermediateMappingsValue.asString();
 
+				if (!Patterns.VALID_INTERMEDIATE.matcher(mappings).matches()) {
+					throw parseException(intermediateMappingsValue, "intermediate_mappings must be a valid maven coordinate");
+				}
+
 				if (!Arrays.asList(supported_mappings).contains(mappings)) {
 					throw parseException(intermediateMappingsValue, "unknown intermediate mappings");
 				}
-			} else {
-				mappings = "hashed";
 			}
 
 			// Until Loader supports hashed mappings
-			if (mappings.equals("hashed")) {
+			if (mappings.equals("org.quiltmc:hashed")) {
 				throw new ParseException("Oh no! This version of Quilt Loader doesn't support hashed mappings, please update Quilt Loader to use this mod.");
 			}
 
--- a/src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
+++ b/src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
@@ -61,6 +61,7 @@ final class V1ModMetadataReader {
 		Map<String, String> contactInformation = new LinkedHashMap<>();
 		List<ModDependency> depends = new ArrayList<>();
 		List<ModDependency> breaks = new ArrayList<>();
+		String intermediateMappings = "intermediary";
 		Icons icons = null;
 		/* Internal fields */
 		ModLoadType loadType = ModLoadType.IF_REQUIRED;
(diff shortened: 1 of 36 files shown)


src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
```diff
@@ -237,6 +238,21 @@ private static V1ModMetadataImpl readFields(JsonLoaderValue.ObjectImpl root) {
 				}
 			}
 
+			@Nullable
+			JsonLoaderValue intermediateMappingsValue = quiltLoader.get("intermediate_mappings");
+
+			if (intermediateMappingsValue != null) {
```
TheGlitch76, line 244: If it is null, we can assume that the mappings are hashed (because that is the default).

TheGlitch76 (changes requested):
Thanks for contributing! Don't forget that you'll need to PR an update to the quilt.mod.json specification in the RFC repository as well.

src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
TibiNonEst: I'm assuming you mean intermediary instead of hashed here as hashed mappings are not yet supported. I am just going to add an else block to fallback to intermediary mappings so that parsing exceptions still get thrown if a mod contains an unsupported value for `intermediate_mappings`.

src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
```diff
@@ -237,6 +238,24 @@ private static V1ModMetadataImpl readFields(JsonLoaderValue.ObjectImpl root) {
...
+				if (mappings.equals("hashed")) {
+					throw parseException(intermediateMappingsValue, "Oh no! This version of Quilt Loader doesn't support hashed mappings, please update Quilt Loader to use this mod.");
+				} else if (!mappings.equals("intermediary")) {
+					throw parseException(intermediateMappingsValue, "unknown intermediate mappings");
+				}
+
+				intermediateMappings = mappings;
+			} else {
+				// Fallback to default mappings
+				intermediateMappings = "hashed";
```
TheGlitch76, line 256: This needs to happen before checking if the value is hashed, otherwise you're letting through hashed mods that aren't supported!

TibiNonEst: Hmm, I think I was trying to limit the number of checks that had to happen, but you're right. I'll set it up the way it will be when Loader supports hashed and then fail if hashed after.

TibiNonEst:
Of course, this implementation will have to change if the RFC amendment culminates with a different style being chosen, i.e. using maven coordinates as zml suggested.

TheGlitch76 (approved):

AlexIIL (approved):

Issue #53, opened 2022-01-24
Environment stripping improvements - Lambda stripping


LambdAurora:
Sided lambdas have been quite terrible to modders, every month at least one modder falls under its wrath.

But what is actually the issue?

Let me pull up a very simple example code:
```java
@Environment(EnvType.CLIENT)
public void hello() {
    this.hi(client -> client.player.sendMessage(new LiteralText("Hi!"), false));
}

@Environment(EnvType.CLIENT)
public void hi(ClientConsumer toDo) {
    toDo.consume(MinecraftClient.getInstance());
}

@Environment(EnvType.CLIENT)
public interface ClientConsumer {
    void consume(MinecraftClient client);
}
```

And let's consider this is in a common class, when loaded on a dedicated server 2 methods will be stripped: `hello` and `hi`.
BUT WAIT! There isn't just 2 methods, there's 3! The lambda is another method too, but wait, it doesn't have the annotation, meaning it won't get stripped.

When loading the class, the server will crash due to `ClientConsumer` not existing, which will be quite unintuitive to modders unaware of this specific behavior.

The loader should strip lambda methods that are generated from a stripped method, it's not trivial but it would resolve a lot of seemingly "random" crashes.

Now there's the argument of "you should not use sided methods, you should separate your client code"! Which I partially agree, in some cases separating a feature to the opposite part of the codebase and have all the remaining code in another part will make understanding and maintaining the feature more difficult, sided methods are very practical in that case, they should not be the "always" go-to, but they should not be straight out rejected from designs.
Fixing this will make codebases significantly more stable when those kind of methods are involved.

AlexIIL:
> When loading the class, the server will crash due to ClientConsumer not existing

I can't reproduce that with the example given - even if I remove the `
@Environment(EnvType.CLIENT)` from both the `hello` and `hi` methods. Can you link to a crash report where this issue actually occurs?

AlexIIL:
Actually, I suppose I was being a bit pedantic with my previous comment - the example given can actually be loaded safely on the server, but the following can't:

```java
public void hello() {
        this.hi(client -> {
            ClientWorld cw = client.world;
            World w = cw; // This requires the java verifier loads both ClientWorld and World to make sure you can perform this implicit upcast
            w.random.nextBoolean();
        });
    }
```

AlexIIL:
This has now been implemented as of 0.17.5-beta.8 - although a full release should follow in a couple of days.

[...]

PR #63, opened 2022-04-15
Add quilt's environment annotations, and deprecate fabric's.

AlexIIL:
Implements #53.

This uses specialized annotations for each side (since there are only two this shouldn't be a problem) rather than using the `EnvType` enum. This also means the names are shorter:

```java
@Environment(EnvType.CLIENT)
@Environment(EnvType.SERVER)
@EnvironmentInterface(value = EnvType.CLIENT, itf = IRenderable.class)
```
now becomes:
```java
@ClientOnly
@DedicatedServerOnly
class MyType implements @ClientOnly IRenderable
```

Both `@ClientOnly` and `@DedicatedServerOnly` also have a `stripLambdas` property (defaulting to true), which controls whether or not to also remove methods referenced in only those methods via lambdas. (This allows mods to disable lambda stripping if they don't want it for a particular method, although I'm not sure *why*).

This leaves only a single non-deprecated class in `net.fabricmc.api`: `EnvType`, which this PR doesn't touch.
Fabric's annotations are *not* processed by the new lambda-stripping functionality, to preserve compatibility.


halotroop2288:
How would you implement side-specific behavior with this?

The old code would have been something like this:
```java
switch (FabricLoader.getInstance().getEnvironment()) {
	EnvType.CLIENT -> LOGGER.info("We're on the client");
	EnvType.SERVER -> LOGGER.info("Hello server");
}

**Comment by triphora** [comment]
@halotroop2288 As mentioned in the OP, `EnvType` is not touched. So your example would be:
```java
switch (MinecraftQuiltLoader.getEnvironmentType()) {
	EnvType.CLIENT -> LOGGER.info("We're on the client");
	EnvType.SERVER -> LOGGER.info("Hello server");
}
```
as it is currently and will still be after this PR.

AlexIIL:
Leo brought up a useful suggestion on discord:

> Leo, Agent of Cursed — Today at 10:33
man, it's too bad TYPE_USE annotations are weird
we could've had
public class ChestBlockEntity implements @ClientOnly ChestAnimationProgress

So, I'm going to add them since it makes a lot more sense to be able to do that than declare the interface twice. I'm not sure if this means I should remove `@ClientOnlyInterface` as well.

AlexIIL:
LambdAurora brought up another useful suggestion:

> Question then
If we have a client only package
Does annotating the package will strip all its classes?

Loader doesn't do this yet - but it probably should, so I'm going to try adding them to this PR too.

AlexIIL:
As of https://github.com/QuiltMC/quilt-loader/pull/63/commits/a2e988759e4fb29ed6e7a5bc56fdf9889e756af8 this now strips annotated interfaces. For example:

```java
@ClientOnlyInterface(IRenderable.class)
public class CentrifugeBlockEntity implements IRenderable {

}
```
can now be written as:
```java
public class CentrifugeBlockEntity implements @ClientOnly IRenderable {

}
```

AlexIIL:
Okay, annotating a `package-info` class with `@ClientOnly` or `@DedicatedServerOnly` now works.

I'm not sure if I should remove `@ClientOnlyInterface`, `@ClientOnlyInterfaces`, `@DedicatedServerOnlyInterface` and `@DedicatedServerOnlyInterfaces`, since those just add clutter at this point?

PR #67, opened 2022-04-20
Fix: Update asm to 9.3 to fix crash caused by LWJGL 3.3.1

ImperatorStorm:
Without this patch, Minecraft>=22w16a crashes as asm<9.3 doesn't support Java 19, and LWJGL 3.3.1 includes Java 19 classes.


AlexIIL:
Thanks!

PR #69, opened 2022-04-21
Sort modlist in logs alphabetically (#68)

### LINKED ISSUE #68: Printed mod list is not alphabetically sorted.
When starting up the mod loader prints the list of mods that is being loaded.
This list is currently not alphabetically sorted making user support much more difficult.

NoahvdAa:
We do a lil' streamin'

Only sorts the list used in the for loop since I wasn't sure whether I should change the order of the actual mod candidate list (because of dependencies).

Closes #68.


LambdAurora (approved):

AlexIIL:
I can't see why we wouldn't want this - and yes, we don't want to sort mods during loading, only printing. (It *might* be useful to have the actual order that mods are loaded somewhere, but that can probably be added separately).

AlexIIL:
Thanks for contributing this!

PR #71, opened 2022-04-24
Make sure mod id doesn't contain : when using mavenGroup:id format.

NinjaPhenix:
Basically using the following would try to look up `org.quiltmc::quilt_loader` rather than `org.quiltmc:quilt_loader`

```
"depends": [
      "org.quiltmc:quilt_loader",
]
```

Example error message:
image


LambdAurora (approved):

NinjaPhenix:
Added some tests, although this code is unlikely to break.

AlexIIL (approved):

AlexIIL:
Looks good, thanks!

PR #74, opened 2022-04-27
Properly select mods in the mods folder over nested jars

TheGlitch76:
I'm 99% sure this is the right place for this fix, but since this isn't my domain of code at all I want to make sure I did it right.

This changes behavior to match expectations--if you place a mod in the mods folder, then that version will now always be selected over a JiJ'd mod with the same ID.


AlexIIL:
Implemented by https://github.com/QuiltMC/quilt-loader/commit/5deae3cf71625303c245dcc54b2a8fa514a7dfb2.

PR #78, opened 2022-05-06
Keep Mixin 0.9.2 compatible mods first in the load order

TibiNonEst:
This PR is a temporary fix for https://github.com/FabricMC/Mixin/issues/89. Forcing 0.9.2-compatible mods to be loaded first should stop local variables from leaking into mods using incompatible versions of Mixin.


TheGlitch76 (approved):

AlexIIL (approved):
Looks ok to me!

PR #83, opened 2022-05-24
Begin limited support of Fabric implementation details

TheGlitch76:
This is a part of my current project to improve the fabric/quilt separation, and increase compatibility with Fabric mods. Because these changes are very noisy (and might, in some cases, include breaking changes to Quilt APIs), I will be separating it into several PRs.

This PR contains:
- Superficial separation of Fabric/Quilt in our codebase--Fabric classes are in a different folder, but can still be seen and used from the Quilt folders.
- Support for most of the `launch` and all of the `language` packages of both `net.fabricmc.loader.impl` and the legacy `net.fabricmc.loader`
- Some support for basically free-to-implement impl details (with more to be added in future PRs).
- Some more impl details as needed to support major mods such as Fabric Language Kotlin

Future impl details of Fabric will be added on an as-needed basis; if we find a mod in the wild that uses a class/method, we can try to implement it.


OroArmor (commented):
`src/fabric` should be `src/fabric/java`

CheaterCodes:
ius there a reason you split fabric into api, impl and legacy as well? It appears that it already was somewhat separated anyway by package structure.

src/fabric/impl/java/net/fabricmc/loader/impl/launch/MappingConfiguration.java
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-package net.fabricmc.loader.launch.common;
+package net.fabricmc.loader.impl.launch;
```
violet-sh, line 17: I think it might be nice to de-dupe this with Quilt's `MappingConfiguration` if possible. This can be done by removing this file and just using Quilt's version, but if this file is still wanted we can always delegate to the other. I just think it's preferable to not have magic strings like the target namespace in multiple places.

src/main/java/org/quiltmc/loader/impl/launch/knot/MixinServiceKnot.java
```diff
@@ -102,7 +102,7 @@ public Class<?> findAgentClass(String name, boolean initialize) throws ClassNotF
 
 	@Override
 	public String getName() {
-		return FabricLauncherBase.getLauncher() instanceof Knot ? "Knot/Fabric" : "Launchwrapper/Fabric";
+		return QuiltLauncherBase.getLauncher() instanceof Knot ? "Knot/Fabric" : "Launchwrapper/Fabric";
```
violet-sh: These should probably be changed to `/Quilt`.

src/main/java/org/quiltmc/loader/impl/entrypoint/EntrypointStorage.java
```diff
@@ -38,8 +38,8 @@ interface Entry {
 
 	@SuppressWarnings("deprecation")
 	private static class OldEntry implements Entry {
-		private static final org.quiltmc.loader.impl.language.LanguageAdapter.Options options = org.quiltmc.loader.impl.language.LanguageAdapter.Options.Builder.create()
-			.missingSuperclassBehaviour(org.quiltmc.loader.impl.language.LanguageAdapter.MissingSuperclassBehavior.RETURN_NULL)
+		private static final net.fabricmc.loader.language.LanguageAdapter.Options options = net.fabricmc.loader.language.LanguageAdapter.Options.Builder.create()
+			.missingSuperclassBehaviour(net.fabricmc.loader.language.LanguageAdapter.MissingSuperclassBehavior.RETURN_NULL)
```
violet-sh, line 42: Is there a reason this got moved back to Fabric's implementation?  The Quilt one is still there and this doesn't necessarily seem in the vein of Quilt/Fabric separation, and the only place I can see `EntrypointStorage` used is `QuiltLoaderImpl`.

src/main/java/org/quiltmc/loader/impl/entrypoint/EntrypointStorage.java
```diff
@@ -62,7 +62,7 @@ public String toString() {
 		@Override
 		public <T> T getOrCreate(Class<T> type) throws Exception {
 			if (object == null) {
-				org.quiltmc.loader.impl.language.LanguageAdapter adapter = (org.quiltmc.loader.impl.language.LanguageAdapter) Class.forName(languageAdapter, true, FabricLauncherBase.getLauncher().getTargetClassLoader()).getConstructor().newInstance();
+				net.fabricmc.loader.language.LanguageAdapter adapter = (net.fabricmc.loader.language.LanguageAdapter) Class.forName(languageAdapter, true, QuiltLauncherBase.getLauncher().getTargetClassLoader()).getConstructor().newInstance();
```
violet-sh, line 65: See comment above.

minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/launchwrapper/QuiltClassTransformer.java.old
```diff
@@ -18,7 +18,7 @@ package org.quiltmc.loader.impl.game.minecraft.launchwrapper;
 
 import net.minecraft.launchwrapper.IClassTransformer;
 
-import net.fabricmc.api.EnvType;
+import net.fabricmc.EnvType;
```
violet-sh: I'm assuming the `.old` extension means this isn't to be used, but even so changing the `EnvType` package doesn't make sense. It's still in the `api` package so even if this isn't used it doesn't make sense to change it.

*This applies to all files in `launchwrapper`.*

violet-sh (commented):
Looks good overall! Just a few small things of note.

minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/launchwrapper/QuiltClassTransformer.java.old
TheGlitch76: this was from a bad refactor, i forgot to revert the changes. IJ likes to edit non-code files sometimes, but only sometimes

src/main/java/org/quiltmc/loader/impl/entrypoint/EntrypointStorage.java
TheGlitch76: This is the "old" entry, for old-style language adapters. It exists *only* to support old language adapters from Fabric, so this implementation is basically useless in its current state. Why would someone want to use the old style language adapters if they are writing a Quilt mod?

violet-sh: Hmm, okay I see what you mean makes sense.

TheGlitch76:
> [is] there a reason you split fabric into api, impl and legacy as well? It appears that it already was somewhat separated anyway by package structure.

The separation between API and legacy API is confusing, because legacy API is `net.fabricmc.loader` (no impl/api package). This is the organization that I found I made the least mistakes with.

PR #92, opened 2022-06-19
Fix multiline comment spacing on TOML configs

EnnuiL:
This PR fixes multiline comments on TOML not having the one space padding on their other lines

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/config/NightConfigSerializer.java
+++ b/src/main/java/org/quiltmc/loader/impl/config/NightConfigSerializer.java
@@ -111,7 +111,9 @@ public final class NightConfigSerializer<C extends CommentedConfig> implements S
 
 			if (node.hasMetadata(Comment.TYPE)) {
 				for (String string : node.metadata(Comment.TYPE)) {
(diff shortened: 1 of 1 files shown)


Haven-King:
This is something that should probably be done in the annotation processor in QuiltConfig itself. I would also like to point at that you can use multiple `@Comment` annotations for multiline comments.

TheGlitch76:
With multi-line strings in Java 17 it doesn't make sense to require splitting into multiple annotations

Haven-King:
> With multi-line strings in Java 17 it doesn't make sense to require splitting into multiple annotations

Agreed, the functionality in this PR makes sense. I just thought it was worth mentioning that there is an existing solution for multiline comments.

EnnuiL:
Yeah, I made this PR so that `@Comment` played nicer with Java's multiline strings; I did strumble on the possibility of using multiple `@Comment` while making this PR, but yeah, I prefer the triple double-quotes

Haven-King:
https://github.com/QuiltMC/quilt-config/pull/3 succeeds this pull request.

PR #102, opened 2022-07-25
Version Refactor

TheGlitch76:
This is a full refactor of how Quilt Loader represents Versions, mostly focused on reverting poor decisions (like 3-component semver) and bringing us up-to-date with the newest Fabric APIs (VersionInterval). **Almost all API touched by this PR is a breaking change.**

This PR will allow:
- Better representation of more kinds of versions than the old API
- Error messages during mod solving that can suggest a certain range of versions to install, even with complex dependencies.

Definitely not my cleanest work; reviews + suggestions to improve are welcome


hYdos:
This is missing the original commits with the original authors no? Why would you leave those out if the code isnt yours? This isnt an attack or harassment campaign, its an genuine question

TheGlitch76:
We follow all of our obligations under the Apache 2.0 license to the best of our ability. Preserving commit history is, notably, not one of those obligations.

MartyAlana:
> We follow all of our obligations under the Apache 2.0 license to the best of our ability. Preserving commit history is, notably, not one of those obligations.

What about not just doing the minimum and instead having enough ethics to allow them to get recognized for writing code which is in your loader?

Leo40Git (approved):
LGTM!

src/test/java/org/quiltmc/test/VersionComparisonTests.java
```diff
@@ -86,68 +85,68 @@ public void testVersionComparisions() throws IOException, VersionFormatException
...
+//		SemanticVersionImpl version = SemanticVersionImpl.of(versionString);
+//		Assertions.assertTrue(constraint.matches(version), constraint + " should match " + versionString);
+//	}
+//
+//	private static void assertNotMatching(VersionConstraintImpl constraint, String versionString)
+//		throws VersionFormatException {
+//
+//		SemanticVersionImpl version = SemanticVersionImpl.of(versionString);
+//		Assertions.assertFalse(constraint.matches(version), constraint + " should not match " + versionString);
+//	}
```
AlexIIL, line 150: Did these tests fail, or do they need re-doing?

src/main/java/org/quiltmc/loader/impl/metadata/DependencyOverrides.java
```diff
@@ -1,281 +0,0 @@
...
-
-		static final Operation[] VALUES = values();
-
-		final String operator;
-
-		Operation(String operator) {
-			this.operator = operator;
-		}
-	}
-}
```
AlexIIL, line 281: Why was this class deleted? I know I haven't implemented them yet, but we probably want to keep it around for when it does get implemented. (Or was it moved but I just missed where it went?)

AlexIIL (changes requested):
This looks good, except for dependency override code being removed?

AlexIIL:
Merged (with major edits) in #117

PR #117, opened 2022-09-28
Loader plugins: Stage 1

AlexIIL:
~~*This is a draft, since we're not ready to merge yet.~~

Before this gets merged we'll also need to merge develop into release, since we will probably want to support pre-plugins while we're ironing out bugs in plugins.

I'm primarily opening this now to more easily see the absolute changes that this makes against pre-plugins.

----------------------

Loader Plugins Stage 1 *(this PR)* means that all mod loading, mod solving, the core plugin loop, and error handling are finished. In addition both quilt and fabric mods are loaded by their respective plugins, and those plugins are part of quilt-loader itself. External (additional) plugins can't be loaded yet, however the api for them should be very nearly ready.

Future stages *(not in this PR)* will bring:
- Loading plugins from mods (sooner rather than later)
- Unloading mismatched plugin versions (hopefully, this can get quite complex though).
- Multi-thread support for plugin tasks
- A file replacement & moving system, to allow plugins that update mods to work well in loader itself.
- A proper abstract gui system, to allow said updater plugins to ask the user about what they're doing before doing it. (And for dep downloading in the future)
- Tentative load option support (for dep downloading, the core isn't not fully implemented ATM)


AlexIIL:
Additional changes done by this PR:

- Added entrypoint hook for 1.15 compatibility with Not Enough Crashes
- Added legacy `net.fabricmc.loader.util.version.VersionParsingException`, and made `net.fabricmc.loader.api.VersionParsingException` extend it.
- Added new VersionInterval and VersionRange apis, based on the ones added to fabric-loader
    - This is basically just https://github.com/QuiltMC/quilt-loader/pull/102 but done in a less breaking way.
    - This also changes a lot of files.
- Made versions accept different component counts than just 3.
- Implemented a few `net.fabricmc.loader.impl.ModContainerImpl` methods:
    - getInfo()
    - getOriginUrl()
    - getContainingMod()
    - getContainedMods()
- Added legacy `net.fabricmc.loader.entrypoint.minecraft.hooks.EntrypointUtils`
- Added `ModContainerExt` and `ModMetadataExt` for plugins to implement when providing mods, which extend their non-`ext` conterparts
    - These contain methods that allow plugins to let quilt-loader implement solving for other plugins
- Added a lot of api classes to `org.quiltmc.loader.api.plugin`, which are only public API for loader plugins.
- Added a simple `Text` interface for loader (and plugins) to use when translating stuff.
    - ~~This should probably be renamed to not collide with minecraft's `Text` class.~~ *renamed in https://github.com/QuiltMC/quilt-loader/pull/117/commits/9ba28017b9f24cee4e511ccc24e6bb9bef809e69 to `QuiltLoaderText`*
- Added a config class to store options for loader. Currently this is not written out and read.
  This contains the following options:
    - `loadSubFolders`: (true ATM)
        - Allows mods to be loaded from sub-folders inside `/mods` (provided the folder names don't start with a digit or `>`, `<`, `=`)
        - Folders that contain a file named `quilt_loader_ignored` will be skipped, and no mods (or sub-folders) will be searched for mods.
    - `restrictGameVersions`: (true ATM) allows mods to be loaded from sub-folders inside `/mods` that match the game version (or are prefixed with `>`, `<`, or `=` and match the interval)
    - `innerZipLoadType` needs to be removed after mod file loading gets optimised.
    - `outerZipLoadType` needs to be removed after mod file loading gets optimised.
    - `singleThreadedLoading`: (true ATM): forces all mod loading operations onto the main thread. This makes debugging a LOT easier.
- Added a `<game-dir>/.quilt/transform-cache.zip` file which contains the transformed (and remapped) class files for every mod that needs transforming/remapping.
    - This is generated based on the SHA-256 hash of the mod file sources.
- Vastly improved error display for some solver errors.
- The mod table now includes the plugin which loaded the mod as well as the source path for the mod (going up through each jar-in-jar)
- Language adapters are now initialised when an entrypoint which uses them is first called.
- Changed mixin config handling to be prefixed with `#modid:` to allow mods to use non-globally unique mixin config names
- Added a few more icons for the gui.

AlexIIL:
Oh, it turns out I forgot about another change:

* Unknown files in the mods folder now cause quilt-loader to fail to load #126

CheaterCodes (approved):
Obviously I only glanced over most of the code.
Some things I noticed:
There's 3 java.old files left over, that I assume you don't need anymore.
QuiltLoaderImpl has large chunks of uncommented code that I assume you don't need anymore.
Ther's obviously lots of Fixmes and TODOS and incomplete javadoc.

This is exciting stuff!

PR #135, opened 2022-11-05
Ignore sub-folders that start with a "."

Scrumplex:
Prism Launcher stores mod metadata in `<instance root>/mods/.index`. Quilt tries to load the `.toml` files in there as mods.
I would propose to just ignore directories that start with a dot.

With this PR:
Ignored .index with this PR

Without this PR:
Quilt trying to load TOML files

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
@@ -1585,6 +1585,11 @@ public class QuiltPluginManagerImpl implements QuiltPluginManager {
 							return FileVisitResult.SKIP_SUBTREE;
 						}
 
+						if (name.startsWith(".")) {  // ignore dot-folders
+							node.subIcon(node.manager().iconDisabled());
+							return FileVisitResult.SKIP_SUBTREE;
+						}
+
 						if (Files.exists(dir.resolve("quilt_loader_ignored"))) {
 							node.subIcon(node.manager().iconDisabled());
 							node.addChild(QuiltLoaderText.translate("warn.sub_folder_ignored"))


src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
```diff
@@ -1585,6 +1585,12 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
 							return FileVisitResult.SKIP_SUBTREE;
 						}
```
triphora, line 1586: This can be removed now

src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
```diff
@@ -1585,6 +1585,12 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
...
 
+						if (name.startsWith(".")) {  // ignore dot-folders
+							System.out.println(name);
```
triphora: Debug line spotted!

triphora (changes requested):

src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
Scrumplex: I don't know what you are talking about :trollface:

src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
AlexIIL: No? This check only occurs at the *end* of the folder name - although they could probably be merged instead.

triphora: Oops, yes, my mind misread "ends" as "starts" for whatever reason

AlexIIL:
Thanks for pointing this out! I'm happy to merge this as-is if you don't have anything else you want to push?

Scrumplex:
I thought about doing this for files as well, but I don't know if it makes sense for those.

AlexIIL:
Files are already ignored if they are prefixed with a `.`

https://github.com/QuiltMC/quilt-loader/blob/develop/src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java#L1665-L1670

Scrumplex:
> Files are already ignored if they are prefixed with a `.`

Oh. I didn't check that :eyes:.

I don't have anything else to add. :D

AlexIIL:
Okay, thanks for contributing :)

PR #137, opened 2022-11-06
Implement version-aware sorting for GenericVersionImpl

exaskye:
I heard QLoader was dropping the SemVer requirement. This is great, but the current implementation is a basic lexical sort on the version string. This is not ideal, and is no better than what FLoader does. (For "extended SemVer", which simply has additional segments, it is actually markedly *worse*.)

This PR implements what I call "FlexVer", a SemVer-compatible intuitive comparator for free-form versioning strings as seen in the wild. It's designed to sort versions like people do, rather than attempting to force conformance to a rigid and limited standard.

It works by splitting a string into "numeric" and "non-numeric" parts, and then sorting those lexically, with a few special cases for SemVer compatibility. This could, in theory, entirely replace Quilt's existing SemVer comparator — but I don't suggest this actually be done.

(This mode of operation is similar to how `sort -V` behaves in GNU coreutils.)

I have a test harness I used to verify this implementation is sound, visible in [the standalone implementation](https://github.com/unascribed/FlexVer/tree/trunk/java) — a screenshot of this harness follows:

image

Cyan is numeric, pink is non-numeric, red is SemVer pre-releases, and gray is ignored SemVer-style appendices. As the potential problem space of version number comparisons is infinite, I cannot devise a good way to truly verify this is a completely sound implementation; so, I simply cherry-picked some random examples of versions that I felt would trip up a version comparator, which both do and do not follow SemVer. As such, I also did not include a unit test; I feel a system like this only benefits from regression tests, after basic sanity tests done during development.

It is a non-goal for comparisons between versions of entirely different structures to "make sense" — but nonetheless, it does try its best and will normalize structural differences and fall back to lexical comparison when all else fails. Fundamentally, ***FlexVer does not impose any form of restrictions*** — it will always do its best to compare two strings and *never* throw an exception. If you do pass two completely different versions to it, well....

image

...the results won't necessarily make sense. Garbage in, garbage out.

There is **one known case where FlexVer and SemVer disagree**. It is related to pre-releases, as they are an exception to the intuitive version flow. There is a special case that makes the most common forms of them work, such as `-pre1`, `-rc2`, `-beta.2`, etc. However, this works by looking for a non-numeric component that starts with `-` and is longer than one character. As such, if you have a pre-release starting with a numeral, then FlexVer's normal parsing kicks in, which sorts `1.0.0` and `1.0.0-2` *in the opposite order that SemVer would*. However, I have never seen fully numeric pre-releases like this in the wild; a hyphenated numeric like that is generally meant to represent another "group" of versions — in which case, FlexVer does indeed sort it as expected.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/metadata/qmj/GenericVersionImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/metadata/qmj/GenericVersionImpl.java
@@ -19,6 +19,7 @@ package org.quiltmc.loader.impl.metadata.qmj;
 import org.quiltmc.loader.api.Version;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 public class GenericVersionImpl implements Version.Raw {
@@ -97,14 +98,16 @@ public class GenericVersionImpl implements Version.Raw {
 	 * of characters changes from numeric to non-numeric.
 	 */
 	private static List<Comparable<?>> decompose(String str) {
-		if (str.isEmpty()) return List.of();
+		if (str.isEmpty()) return Collections.emptyList();
 		boolean lastWasNumber = Character.isDigit(str.codePointAt(0));
 		StringBuilder accum = new StringBuilder();
 		List<Comparable<?>> out = new ArrayList<>();
 		// remove appendices
 		int plus = str.lastIndexOf('+');
 		if (plus != -1) str = str.substring(0, plus);
-		for (int cp : str.codePoints().toArray()) {
+		for (int i = 0; i < str.length(); i++) {
+			if (Character.isLowSurrogate(str.charAt(i))) continue;
+			int cp = str.codePointAt(i);
 			boolean number = Character.isDigit(cp);
 			if (number != lastWasNumber) {
 				complete(lastWasNumber, accum, out);
--- a/src/main/java/org/quiltmc/loader/impl/metadata/qmj/GenericVersionImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/metadata/qmj/GenericVersionImpl.java
@@ -53,10 +53,86 @@ public class GenericVersionImpl implements Version.Raw {
 	public int compareTo(Version other) {
 		return compareRaw(raw(), other.raw());
 	}
-
+	// it's difficult to generically deal with comparables - these operations are safe
+	@SuppressWarnings({ "rawtypes", "unchecked" })
 	public static int compareRaw(String a, String b) {
-		// Ideally this would use number-aware comparison
-		// Oh well
-		return a.compareTo(b);
+		// Arrays.compare gives wrong precedence to array length, causing 1.0.1 to sort as older than 1.0.0_01
+		// so, implement it ourselves from scratch
+		List<Comparable<?>> ad = decompose(a);
+		List<Comparable<?>> bd = decompose(b);
+		for (int i = 0; i < Math.max(ad.size(), bd.size()); i++) {
+			Comparable ac = i >= ad.size() ? null : ad.get(i);
+			Comparable bc = i >= bd.size() ? null : bd.get(i);
+			if (typeof(ac) != typeof(bc)) {
+				// Comparables assume the input object is the same type as the object, so we need to
+				// ensure that's true; this will happen in the case of two critically mismatched
+				// versions that contain a symbol where a number is expected - in this case, lexical
+				// is the best we can do
+				ac = ac == null ? null : ac.toString();
+				bc = bc == null ? null : bc.toString();
+			}
+			int c;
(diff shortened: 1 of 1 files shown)


LemmaEOF (approved):

Leo40Git (approved):

exaskye:
Full disclosure: I wrote this in a detached file and added it to QLoader via the web editor. I know it compiles and works, but my drop-in of this into the project missed the imports. :sweat_smile:

exaskye:
Aand apparently I accidentally used new Java API. <.<

That I would not have caught even if I did go to the trouble of importing QLoader into my IDE.

violet-sh (approved):

TheGlitch76 (approved):

sisby-folk (approved):
good writeup
-kapesi

AlexIIL:
Wow, this looks amazing - thanks for PRing this! I'm sadly not ready to merge now (as I'm about to go to sleep) but I'll clone it locally tomorrow and have a look then :)

sisby-folk:
Kapesi: 

Bonus practical problem this solves that we're currently experiencing, that doesn't have anything to do with bespoke alternate versioning schemes at all.

Our mod, [Portable Crafting Standalone](https://modrinth.com/mod/portable-crafting-standalone), has an optional [Inventory Tabs](https://modrinth.com/mod/inventory-tabs-updated/) dependency set to `>=0.6.0-1.19.x`.

The current release of this mod has invalid semver - the version looks okay at a glance, but is missing some expected separators

it's `0.9.beta-1.19.x` - instead of the semver-appropriate `0.9.0-beta+1.19.x`.

Right now, that means that downloading the latest release of both mods will cause a startup crash for dependency resolution, because inventory tabs is an "incompatible" version compared to what PCS is looking for (because it can't be guaranteed to be later or equal to 0.6.0)

With flexver as a fallback, the loader would see one version (the dependency) as valid semver, and one version (the loaded mod) as not semver, and use flexver to make the comparison. Here, `0.9` is determined to be higher than `0.6`, and the rest is history - job solved, no more users saying we have a broken mod.

---

Of course, that already would kinda work with string compare - See, the important part is that "`0.9.beta-1.19.x`" *starts* with valid version numbers that could be compared numerically - and I feel this is very common - when semver fails, often it's because one version ends with something illegal.

But - if Andrew6rant released version `0.11.beta-1.19.x` tomorrow - the old comparator would fail and insist it's earlier than `0.6`, but a flexver fallback would validate the version - The numeric comparison would clearly see `0.11` as being later than `0.6`. 

Perhaps intuitively - "Compares exactly the same as semver until it encounters illegal semver, then does its best" actually solves a lot of problems.

LambdAurora (approved):

exaskye:
Since making this PR, I've written a [specification](https://github.com/unascribed/FlexVer/blob/trunk/SPEC.md) and created a [more robust and idiomatic Java implementation](https://github.com/unascribed/FlexVer/blob/trunk/java/src/main/java/com/unascribed/flexver/FlexVerComparator.java), which is published to Maven. I intentionally kept the implementation in this PR simple and didn't use inner classes. This PR was sort of the catalyst for me finally buckling down and doing what I've been threatening to for years, heh.

Do we want to keep the current implementation in this PR, or would it be better to depend on FlexVer-Java via Gradle and use its API? I'm not sure what the QLoader policy on dependencies is.

AlexIIL:
I was just about to comment with some issues in the original code - but it looks like you've fixed all those in your better implementation!

Quilt-loader mainly doesn't want to depend on large things - since we need to mirror them to our maven, or (even better) shadow-include them in the release jar to prevent mods from accidentally depending on the wrong version. Obviously this isn't much of an issue here since it's just a single class, so I'd be happier shadow-including that library than including a worse version directly in `GenericVersionImpl`.

I do have one question though - why did you remove the usage of `String.codePoints()`? That method (unlike `List.of()` is actually available in java 8.

AlexIIL:
Oh and a bit more of a minor issue - `org.quiltmc.loader.api.Version.compareTo` currently specifies that it compares non-semer using `String.compareTo` - that should probably be changed to point to the FlexVer specification.

Either way I'd be happy to implement all these changes myself - you've already done all of the hard work by defining the spec and writing the impl after all, which I think everyone would rather we use instead of `String.compareTo`.

exaskye:
Hm, I somehow misread the String::codePoints doc as having been introduced in 9 rather than 8. Technically, my implementation is a little more efficient as it doesn't involve a Stream, but oops. It feels a little silly to change it now.

exaskye:
As you're more familiar with the project structure, it is likely better if you take the liberty of adding the dependency via Gradle and updating the docs. We can consider this PR a discussion thread and supersede it with such a change.

AlexIIL:
Ok, I'll do that now.

exaskye:
:tada:

PR #161, opened 2022-11-29
Begin the process of hiding loader internals from mods

AlexIIL:
Since mods have previously been able to access loader's internals (both by reflection and compile code against it) this doesn't restrict access to any class which existed before loader `0.18.0`, and we probably shouldn't in the future - hence why we can't multi-release-jar a module-info.java file and be done with it.

Instead this adds a new annotation:
```java
/** Indicates that the specified quilt-loader class is internal, and not loadable by mods.
 * <p>
 * Quilt-loader plugin APIs are annotated with this, but use {@link QuiltLoaderInternalType#PLUGIN_API} as their value -
 * which indicates that it's the only quilt-loader classes that loader plugins may use.
 * <p>
 * By default this marks the class as "legacy_exposed", which doesn't throw an exception when trying to access it. */
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.TYPE)
public @interface QuiltLoaderInternal {

	/** Controls how "internal" the class is. */
	QuiltLoaderInternalType value();
}
```
and it's corresponding enum value:
```java

@QuiltLoaderInternal(QuiltLoaderInternalType.NEW_INTERNAL)
public enum QuiltLoaderInternalType {

	/** Indicates the class is both legacy, and was originally considered "api" - so no warnings should be printed on
	 * callers. */
	LEGACY_NO_WARN,

	/** Indicates the class is legacy, and not considered "api" at any point - so warnings should be printed when
	 * callers try to access it. */
	LEGACY_EXPOSED,

	/** Permits loader plugins to access the class, but not mods. */
	PLUGIN_API,

	/** Indicates the class has been added since 0.18.0, and so both mods and plugins aren't allowed to use them. */
	NEW_INTERNAL;
}
```

This also annotated all existing internal classes with the annotation.


AlexIIL:
TODO: add the relevant annotations to `QuiltLoaderInternalType` and `QuiltLoaderInternal` - previously I thought you couldn't do this (I got a compile error the first time that I tried - but not the second time, so perhaps something else prevented it from working originally).

src/main/java/org/quiltmc/loader/impl/launch/knot/KnotClassDelegate.java
```diff
@@ -129,6 +132,80 @@ private IMixinTransformer getMixinTransformer() {
...
+				}
+				case NEW_INTERNAL:
+				default: {
+					throw new IllegalQuiltInternalAccessError("! Quilt-loader internal " + c
+						+ "\nPlease don't use this, instead ask us to declare a new public API that we can guarantee backwards compatibility for!");
+				}
+				case PLUGIN_API: {
+					throw new IllegalQuiltInternalAccessError("! Quilt-loader plugin-only internal " + c
+						+ "\nPlease don't use this, instead ask us to declare a new public API that we can guarantee backwards compatibility for!");
+				}
```
TheGlitch76: Something else to consider is that often internal API does have a public equivalent, but they just missed it.

AlexIIL: That's a good point - perhaps the annotation could include a description of what might be a replacement?

AlexIIL: On second thought, perhaps a separate annotation that could be applied to methods and fields too, to be checked in... something else?

PR #165, opened 2022-12-05
Remove semver warnings

exaskye:
With the advent of 4e4951221eaaceb1962eaa4297921e7a68326ffd (as discussed in #137) these warnings are now entirely pointless spam left over from FLoader. As QLoader now has a fully featured generic version comparator (and the Fabric Version API is just a facade for Quilt's), printing this because someone had the audacity to install Create or one of dozens of other mods that don't strictly follow semver is pointless and annoying.


AlexIIL:
Yes, thanks!

PR #183, opened 2022-12-29
Initial chasm support

AlexIIL:
This requires the user:

- Enable chasm support with `-Dloader.experimental.enable_chasm=true`
- Include a chasm jar in their classpath (manually, it's not added by quilt-loader)

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/discovery/RuntimeModRemapper.java
+++ b/src/main/java/org/quiltmc/loader/impl/discovery/RuntimeModRemapper.java
@@ -62,6 +62,15 @@ public final class RuntimeModRemapper {
 				.collect(Collectors.toList());
 
 		if (modsToRemap.isEmpty()) {
+			// Nothing to remap, but still copy over:
+			// - class files for all
+			// - resources for everything not loaded from a folder on the default FS (since that can change at any time)
+			for (ModLoadOption mod : modList) {
+				if (!mod.needsChasmTransforming()) {
+					continue;
+				}
+				Path resourceRoot = mod.resourceRoot();
+			}
 			return;
 		}
 
--- a/src/main/java/org/quiltmc/loader/impl/discovery/RuntimeModRemapper.java
+++ b/src/main/java/org/quiltmc/loader/impl/discovery/RuntimeModRemapper.java
@@ -61,16 +61,32 @@ public final class RuntimeModRemapper {
 				.filter(modLoadOption -> modLoadOption.namespaceMappingFrom() != null)
 				.collect(Collectors.toList());
 
-		if (modsToRemap.isEmpty()) {
-			// Nothing to remap, but still copy over:
-			// - class files for all
-			// - resources for everything not loaded from a folder on the default FS (since that can change at any time)
-			for (ModLoadOption mod : modList) {
-				if (!mod.needsChasmTransforming()) {
-					continue;
+		// Copy everything that's not in the mods to remap list
+		for (ModLoadOption mod : modList) {
+			if (mod.namespaceMappingFrom() == null && mod.needsChasmTransforming() && !"quilt_loader".equals(mod.id())) {
+				Path modSrc = mod.resourceRoot();
+				Path modDst = cache.resolve(mod.id());
+				try {
+					Files.walk(modSrc).forEach(path -> {
+						if (!Files.isRegularFile(path)) {
+							return;
+						}
+						Path sub = modSrc.relativize(path);
+						Path dst = modDst.resolve(sub.toString());
+						try {
+							Files.createDirectories(dst.getParent());
+							Files.copy(path, dst);
+						} catch (IOException e) {
+							throw new Error(e);
+						}
+					});
+				} catch (IOException io) {
+					throw new Error(io);
 				}
-				Path resourceRoot = mod.resourceRoot();
 			}
+		}
+
+		if (modsToRemap.isEmpty()) {
 			return;
 		}
 
--- a/src/main/java/org/quiltmc/loader/impl/discovery/RuntimeModRemapper.java
+++ b/src/main/java/org/quiltmc/loader/impl/discovery/RuntimeModRemapper.java
@@ -38,6 +38,7 @@ import java.util.stream.Collectors;
 
 import org.objectweb.asm.commons.Remapper;
 import org.quiltmc.loader.api.plugin.solver.ModLoadOption;
+import org.quiltmc.loader.impl.QuiltLoaderImpl;
 import org.quiltmc.loader.impl.filesystem.QuiltMemoryFileSystem;
 import org.quiltmc.loader.impl.launch.common.QuiltLauncher;
 import org.quiltmc.loader.impl.launch.common.QuiltLauncherBase;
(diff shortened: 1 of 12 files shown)


src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
```diff
@@ -0,0 +1,243 @@
...
+					}
+				}
+				return null;
+			}
+
+			@Override
+			public boolean isInterface(String className) {
+				ClassReader cr = new ClassReader(readFile(LoaderUtil.getClassFileName(className)));
+				ClassInfoGrabber namer = new ClassInfoGrabber();
+				cr.accept(namer, ClassReader.SKIP_CODE | ClassReader.SKIP_DEBUG | ClassReader.SKIP_FRAMES);
```
CheaterCodes: I think `cr.accept()` parses the entire class. `cr.getAccess()` should give the same result, but much cheaper.
Similarly, for the remaining occurances of accept, I *think* `cr.getInterfaces()`, `cr.getClassName()` and `cr.getSuperName()` are significantly cheaper for larger classes.

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
```diff
@@ -0,0 +1,243 @@
...
+		for (ModLoadOption mod : modList) {
+			Path modPath = root.resolve(mod.id());
+			if (!Files.exists(modPath)) {
+				continue;
+			}
+			Path chasmRoot = modPath.resolve("chasm_transformers");
+			Files.walkFileTree(modPath, new SimpleFileVisitor<Path>() {
+				@Override
+				public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+					if (file.getFileName().toString().endsWith(".class")) {
```
CheaterCodes, line 206: Do we support multi-version jars? If so, are classes for other versions already stripped? Otherwise this will cause problems.

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
```diff
@@ -0,0 +1,243 @@
...
+				return namer.superName;
+			}
+		});
+
+		for (ModLoadOption mod : modList) {
+			Path modPath = root.resolve(mod.id());
+			if (!Files.exists(modPath)) {
+				continue;
+			}
+			Path chasmRoot = modPath.resolve("chasm_transformers");
```
CheaterCodes: I guess this is where we decide on where transformers should be put. My plan was to use `org/quiltmc/chasm/transformers`, but I suppose it doesn't really matter and we can decide at a later date.

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
```diff
@@ -0,0 +1,243 @@
...
+						Node node = Node.parse(file);
+						Transformer transformer = new ChasmLangTransformer(chasmId, node, chasm.getContext());
+						chasm.addTransformer(transformer);
+					}
+					return FileVisitResult.CONTINUE;
+				}
+			});
+		}
+
+		List<ClassData> changed = chasm.process(true);
```
CheaterCodes, line 231: Note that this will not tell you if a class got completely removed iirc. Might be something that needs to be changed, but not sure what the API would look like.

src/main/java/org/quiltmc/loader/impl/discovery/RuntimeModRemapper.java
```diff
@@ -61,6 +61,31 @@ public static void remap(Path cache, List<ModLoadOption> modList) {
 				.filter(modLoadOption -> modLoadOption.namespaceMappingFrom() != null)
 				.collect(Collectors.toList());
 
+		// Copy everything that's not in the mods to remap list
+		for (ModLoadOption mod : modList) {
+			if (mod.namespaceMappingFrom() == null && mod.needsChasmTransforming() && !"quilt_loader".equals(mod.id())) {
```
CheaterCodes: This should probably use the same `MOD_ID` constant you use in `QuiltLoaderImpl`

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
AlexIIL: I wasn't aware of those methods - thanks!

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
AlexIIL: I'm not sure, but I don't think this is the right place to add support - processing jars before chasm to only have a single class file would probably be a good idea. Oh, and validating that the package + class name == file path + ".class" is probably a good idea before passing the classes to chasm

CheaterCodes: Yeah I agree that this should be done earlier, just wanted to point it out since I ran into that problem once

CheaterCodes (approved):

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
AlexIIL: Ah, I didn't think about this in terms of URLs & domains - it makes a lot of sense to use that instead.

CheaterCodes: As mentioned in discord, another option would be to have the root path be user-defined in QMJ, since people might find it strange to use someone elses namespace.

AlexIIL: Yeah, that's definitely a sensible option - but I'd rather not touch QMJ when we're still in the experimental stage.

AlexIIL: Okay, I take that back - now chasm files are expected to be loaded from folders specified by `experimental_chasm_transformers` in the root of the `quilt.mod.json`.

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
```diff
@@ -173,12 +208,20 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
-						String chasmId = mod.id() + ":" + chasmRoot.relativize(chasmRoot).toString();
-						Log.info(LogCategory.GENERAL, "Found chasm transformer: '" + chasmId + "'");
-						Node node = Node.parse(file);
-						Transformer transformer = new ChasmLangTransformer(chasmId, node, chasm.getContext());
-						chasm.addTransformer(transformer);
+					} else if (file.getFileName().toString().endsWith(".chasm")) {
+						for (Path chasmRoot : chasmRoots) {
+							if (file.startsWith(chasmRoot)) {
+								String chasmId = mod.id() + ":" + chasmRoot.relativize(file).toString();
+								if (chasmId.endsWith(".chasm")) {
```
CheaterCodes, line 215: isn't this always true? It's checked a few lines earlier

CheaterCodes: not sure why this review was duplicated. Thanks GitHub I guess.

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
CheaterCodes: It seems like other people preferred this as well, so sounds good.

src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
CheaterCodes: Just letting you know that I opened an issue here: https://github.com/QuiltMC/chasm/issues/96
This will be a breaking change to this part of the code

PR #188, opened 2023-01-04
Add directory support to GameTransformer

Ampflower:
Found by attempting to run Quilt in a development environment with a GameProvider targeting what was being developed.

Essentially lets the GameTransformer be able to handle directory code sources, which should also help in cases that the end user for any reason unjars Minecraft then tries to run it with Quilt.

It'd also be of interest to have this backported to 0.17 to have a stable release with this patch.


AlexIIL (approved):
This looks reasonable - it might be a good idea to rename the `gameJar` parameter to `gameFile` since it can also be a folder.

Ampflower:
Hm, one other alternative would be using `gamePath` or similar? Since if it is a directory, it's not exactly a file.

AlexIIL:
Yes, `gamePath` would make more sense

PR #190, opened 2023-01-05
Add dependency metadata to the maven publication.

Ampflower:
Provides valid dependency declarations in the maven metadata for using Quilt where Loom is otherwise not required, such as a Discord bot. This primarily negates the need for the end developer from having to specify the same set of dependencies that would've otherwise been automatically specified with `from(components["java"])`.

This approach can be extended to the launcher manifests if desired, likely reducing the overall code and maintenance required for adding a new non-shaded dependency.

Note that this only does the `api` dependency and entirely ignores all other declarations (unless `components.api` extends another component.)

The reason why `from(components["java"])` wasn't used is due to the inability to cleanly exclude dependencies from the resulting metadata, leading to unnecessary dependency downloads for stuff that Quilt instead shades in and strips down for its own use.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -353,6 +353,41 @@ publishing {
 //				builtBy copyJson
 //				classifier = "launchwrapper"
 //			}
+
+			// Manually inject the dependencies into the final POM.
+			// This allows for use without Loom in projects that don't need it.
+			//
+			// This is done this way to avoid including dependencies that are
+			// shaded in, as Gradle seems to provide no clean way of excluding
+			// dependencies from the final POM.
+			//
+			// Note: This only uses the `api` configuration to determine what is
+			// fine to include, and does not respect exclusions. A more advanced
+			// implementation that excludes the `include` configuration from
+			// runtime & compile classpaths would likely prove to be better if
+			// Quilt ever needs to have dependencies that aren't to be shaded nor
+			// passed through to the application.
+			pom.withXml {
+				final def dependenciesNode = asNode().appendNode("dependencies")
+
+				for (final def dep : configurations.api.allDependencies) {
+					// Tests for an external dependency.
+					// Provides a guarantee that the needed metadata is present.
+					if (dep instanceof ExternalDependency) {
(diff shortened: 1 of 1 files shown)


AlexIIL:
Thanks! I'm not sure if this breaks anything, but I think the easiest way to find out is to merge this and then wait to see if anyone complains :p

AlexIIL:
This is probably getting reverted since loom would also need updating to handle this - although I'll let @TheGlitch76 have the final say over whether this gets reverted or not.

(Apparently this was tried before and it also didn't work very well).

Ampflower:
Doing some investigating, it seems that due to Mixin, older versions of some libraries, GSON and Guava, are being pulled in via the modApi and modImplementation classpaths without being updated to match the Minecraft classpath. This doesn't have any effect in most standard mod environments; *GSON and Guava both use up-to-date versions and the older versions are correctly shadowed*.

It does seem that QFAPI's usage of Loom and Loader does mean it has uncovered an underlying fault that was otherwise handled by Gradle internals silently, as it so far has been the only one to produce such a fault, behaving as if the modApi classpath was copied over without any bearing of the Minecraft classpath. Adding an up-to-date GSON to the modApi configuration allows it to work normally.

Ampflower:
At this point, I think QFAPI's setup is cursed.
I don't see how testCompileClasspath is missing Minecraft libraries only in its setup. Up-to-date GSON is present in another project using the same version of Loom, Gradle and Loader on all classpaths (aside from the obvious mod* ones). I can only assume how Loom deals with separating client & server for split development environments causes compileClasspath to not transfer correctly into testCompileClasspath and testmodCompileClasspath.

Ampflower:
Update: This seems to entirely be a Loom bug with how `loom.splitEnvironmentSourceSets()` is handled. `testCompileClasspath` and `testRuntimeClasspath` both fail to inherit `compileClasspath` and `runtimeClasspath`, as `./gradlew dependencies` in a single mod project shows that the classpaths aren't correctly populated, and IntelliJ agrees with it.

PR #192, opened 2023-01-06
Fix dependency override removal warning logic being inverted

LostLuma:
Inverts the logic for when a dependency override removal logs a warning, as it currently logs when it did find a dependency to remove, rather than when it did not. I moved the warning code out of the block so this is easier to read, overall.

I've also removed two unused imports as my editor was screaming at me for being unable to find the first one, and it doesn't seem like it was added intentionally.


AlexIIL:
Thanks!

PR #195, opened 2023-01-10
Fix Fabric Loader Api getting Quilt Metadata

hexadecimal233:
This solves some fabric mods try to get authors and returns empty.


Ampflower (commented):
Wasn't.. this already giving the Fabric mod metadata?

This seems like a band-aid around the root issue if this isn't already returning what it should be.

AlexIIL:
Please open an issue describing what the problem is.

PR #205, opened 2023-01-18
Fix findEmbeddedGameProvider searching in Fabric's namespace.

Ampflower:
This is a really minor bug at the end of it, but should be good to catch.


AlexIIL:
Thanks!

PR #206, opened 2023-01-18
Fix crash from getting super of java/lang/Object

Ampflower:
This is a blanket ban of `java/`, given that it's normally not expected to be remapped anyways.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/util/mappings/MixinIntermediaryDevRemapper.java
+++ b/src/main/java/org/quiltmc/loader/impl/util/mappings/MixinIntermediaryDevRemapper.java
@@ -142,6 +142,11 @@ public class MixinIntermediaryDevRemapper extends MixinRemapper {
 			}
 		}
 
+		// Realistically, this can't be remapped.
+		if (owner.startsWith("java/")) {
+			return name;
+		}
+
 		ClassInfo classInfo = ClassInfo.forName(owner);
 
 		if (classInfo == null) { // unknown class?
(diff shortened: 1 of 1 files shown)


Ampflower:
Also, curiously, should this be PR'd into `release` as well, given that this is likely one of the changes 0.14.12 had for dev remapping failure?

src/main/java/org/quiltmc/loader/impl/util/mappings/MixinIntermediaryDevRemapper.java
```diff
@@ -142,6 +142,11 @@ public String mapMethodName(String owner, String name, String desc) {
 			}
 		}
 
+		// Realistically, this can't be remapped.
+		if (owner.startsWith("java/")) {
+			return name;
+		}
+
```
AlexIIL: You could probably move this to the top of the method? (With a null check:p). Other than that this looks good.

AlexIIL (approved):

AlexIIL:
> should this be PR'd into release as well

Yeah, that's a good idea - especially as it's a simple change.

src/main/java/org/quiltmc/loader/impl/util/mappings/MixinIntermediaryDevRemapper.java
Ampflower: Think I placed it after the first `if` to have the HashSet get a hit in case it's already hashed by a previous call, but I could move it to the top as well.

Ampflower: It does seem that the likely hood of it being a hit is practically none if it's directly from Mixin, so, readability wins here.

Ampflower: image

Tho, it does seem to be less trivial given the fact that it can be null, unless I do an `owner != null` check before comparing strings.

PR #215, opened 2023-01-22
Revert "Add dependency metadata to the maven publication."

AlexIIL:
Reverts QuiltMC/quilt-loader#190, since apparently it's been tried before and it breaks loom


TheGlitch76:
45e76713c415f27f22293e03d7bc67dbddb9c26f

PR #226, opened 2023-01-27
Update Quilt Meta CI

Akarys42:
This commit removes the `refreshMeta` task and replaces it by the new GitHub Actions setup. The manual refresh trigger has also been removed in favor of the one in the action repo.

Draft until the last secret is added to the organisation.


AlexIIL:
Can you have a look at the error I'm getting at the moment when using this? https://github.com/QuiltMC/quilt-loader/actions/runs/4036584305/jobs/6939279366

PR #245, opened 2023-02-21
Fix ClassNotFoundException involving AppletMain

Copetan:
This fixes a bug in which `EntrypointPatch.process()` was pointing to `net/fabricmc/loader/impl/game/minecraft/applet/AppletMain` instead of `org/quiltmc/loader/impl/game/minecraft/applet/AppletMain`, leading to an exception when trying to run Quilt on 1.5.2


AlexIIL (approved):

AlexIIL:
Thanks! I

PR #252, opened 2023-02-27
fix restricted class detection in `KnotClassDelegate`

### LINKED ISSUE #251: Certain legacy server versions crash while trying to load mods
When trying to load a 1.7.2 server with Quilt and a mod, [the game crashes due to a `MixinTargetAlreadyLoadedException`](https://mclo.gs/6MzSDGb). With the help of @SpaceWalkerRS, we tested on other versions and found that the same issue exists in 1.8.9, but not 1.9 nor 1.6.4.

I looked into it more deeply and found that [the log4j2.xml file in 1.7.2](https://mclo.gs/wK4uCb5) is slightly different from [the one in 1.9](https://mclo.gs/4jcBEN9); more specifically, the `packages` attribute in the `Configuration` element lists `net.minecraft` and `com.mojang` in 1.7.2, whereas 1.9 only lists `com.mojang.util`. I guess while trying to find Log4j plugins, the class loader loads every class within the listed packages, in turn loading Minecraft classes earlier than anticipated and breaking Mixin processing. (The server jar started bundling Log4j in 13w39a, any older versions don't have the class loading issue, seemingly because they don't use Log4j.)

While stepping through the code in a debugger, I found that there's a mechanism in [`KnotClassDelegate`](https://github.com/QuiltMC/quilt-loader/blob/develop/src/main/java/org/quiltmc/loader/impl/launch/knot/KnotClassDelegate.java) that restricts certain classes from being loaded, so I think this issue might have something to do with that?

Copetan:
A one line change feels easier than I expected for this, I hope there aren't any unintentional consequences 😓 
Fixes #251

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/gui/QuiltLoaderTextImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/gui/QuiltLoaderTextImpl.java
@@ -16,12 +16,12 @@
 
 package org.quiltmc.loader.impl.gui;
 
+import java.util.IllegalFormatException;
+
 import org.quiltmc.loader.api.gui.QuiltLoaderText;
 import org.quiltmc.loader.impl.plugin.gui.I18n;
 import org.quiltmc.loader.impl.util.QuiltLoaderInternal;
 import org.quiltmc.loader.impl.util.QuiltLoaderInternalType;
-
-import java.util.MissingFormatArgumentException;
 
 @QuiltLoaderInternal(QuiltLoaderInternalType.NEW_INTERNAL)
 public final class QuiltLoaderTextImpl implements QuiltLoaderText {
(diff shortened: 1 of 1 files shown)


TheGlitch76 (changes requested):
It's not clear at all what this magic regex is doing just by looking at it. A few things:
- This should note what it's doing and why (What is this matching for? Classes? why no `$` detection? what problem does this solve?)
- Is a regex really the right solution here? This code compiles a new regex and runs it every single time a class loads, which seems a little excessive to me. Could this be replaced by one or two `startsWith`, `endsWith`, or `contains` tests, or a pre-compiled `Pattern` if necessary?

AlexIIL:
I'm sorry for not responding before.

The current code is broken, since `allowedPrefixes` contains a map of class-path `URL`s (like `quilt.mfs://transformed-mod-minecraft.i0/`), but the current code tests for the full class URL (like `quilt.mfs://transformed-mod-minecraft.i0/net/minecraft/class_15461234.class`) with `Map.get()`, which doesn't work.

As such this PR changes the check to remove the class name suffix to check against the actual URL, which will (presumably) fix the actual bug. (This change also removes the leading `!` since - I assume - the bug occurs to files loaded from zips).

Before this can be merged though this absolutely should be replaced with a precompiled `Pattern`, since compiling regex is quite slow, and classes get loaded a lot.

PR #258, opened 2023-03-12
Wire up loading plugins, for fun.

TheGlitch76:
I'd like to start messing with loader plugins external to Quilt Loader, both for my own amusement and to figure out what kinds of things we might want, find limitations of the current approach, etc.

This PR provides an undocumented way to add a loader plugin from a mod, locked behind a system property. If the system property isn't set, an intentionally vague error message is given. This is intended to prevent people from shipping their own loader plugins just yet (though I know it will happen anyway)

### DIFF
--- a/src/main/java/org/quiltmc/loader/api/plugin/ModMetadataExt.java
+++ b/src/main/java/org/quiltmc/loader/api/plugin/ModMetadataExt.java
@@ -54,9 +54,7 @@ public interface ModMetadataExt extends ModMetadata, ModMetadataToBeMovedToPlugi
 	}
 
 	@Nullable
-	default ModPlugin plugin() {
-		return null;
-	}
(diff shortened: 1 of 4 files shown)


AlexIIL (approved):

src/main/java/org/quiltmc/loader/api/plugin/ModMetadataExt.java
```diff
@@ -54,9 +54,7 @@ default ModLoadType loadType() {
 	}
 
 	@Nullable
-	default ModPlugin plugin() {
-		return null;
-	}
+	ModPlugin plugin();
```
AlexIIL, line 57: I'd rather this was null by default to be honest actually, since I'm not sure how stable plugins-loading-plugins will end up being?

PR #262, opened 2023-03-14
Add proper (localhost) forked-process gui support

AlexIIL:
This adds a way for mods to display errors and receive button callbacks via `QuiltLoaderGui`, and improves a few misc gui things. This will eventually be used to run plugin code in the swing process, so that plugins can display custom UIs.

This is used to display if there's an error with the `quilt-loader-overrides.json` files:
Screenshot from 2023-03-14 09-22-03
Clicking the `Reload` button after fixing the problem changes the screen to this:
Screenshot from 2023-03-14 09-20-26


TheGlitch76 (approved):
I skimmed and this looks OK to me. Let's go ahead and consider this the start of 0.19.

We should probably try to keep 0.19 mostly focused on GUI improvements, so we don't have a repeat of 0.18.

PR #265, opened 2023-03-19
Ignore duplicate mods of the same version in development environments.

TheGlitch76:
Quilt Loader 0.17 and Fabric Loader does this in all environments, but I don't think we should accept duplicate mods in production.
However, many complicated, multi-mod dev environments (QSL, QFAPI) depend on this behavior. While it is probably possible to fandangle a buildscript which does not put the same mod on the classpath multiple times, this is a much simpler solution that matches generally expected behavior.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
@@ -1927,8 +1927,9 @@ public class QuiltPluginManagerImpl implements QuiltPluginManager {
 
 		PotentialModSet set = modIds.computeIfAbsent(id, k -> new PotentialModSet());
 
-		if (set.byVersionSingles.containsKey(version) && QuiltLoader.isDevelopmentEnvironment()) {
-			Log.debug(LogCategory.SOLVING, String.format("Ignoring duplicate mod %s of the same version %s loaded from %s", id, version, from));
+		ModLoadOption current = set.byVersionSingles.get(version);
+		if (current != null && current.isMandatory() && mod.isMandatory() && current.getClass() == mod.getClass() && QuiltLoader.isDevelopmentEnvironment()) {
+			Log.warn(LogCategory.SOLVING, String.format("Ignoring duplicate mod %s of the same version %s loaded from %s", id, version, from));
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
```diff
@@ -1921,26 +1921,17 @@ private void addModOption(Path file, Map<ModLoadOption, BasePluginContext> map,
...
-		modPaths.put(from, mod);
-		modProviders.put(mod, provider.pluginId());
-		modGuiNodes.put(mod, guiNode);
-
-		guiNode.addChild(QuiltLoaderText.translate("gui.text.id", id));
-		guiNode.addChild(QuiltLoaderText.translate("gui.text.version", version.raw()));
 
 		PotentialModSet set = modIds.computeIfAbsent(id, k -> new PotentialModSet());
+
+		if (set.byVersionSingles.containsKey(version) && QuiltLoader.isDevelopmentEnvironment()) {
```
AlexIIL, line 1930:
```suggestion
		ModLoadOption current = set.byVersionSingles.get(version);
		if (current != null && current.isMandatory() && mod.isMandatory() && QuiltLoader.isDevelopmentEnvironment()) {
```

src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
```diff
@@ -1921,26 +1921,17 @@ private void addModOption(Path file, Map<ModLoadOption, BasePluginContext> map,
...
+			Log.debug(LogCategory.SOLVING, String.format("Ignoring duplicate mod %s of the same version %s loaded from %s", id, version, from));
```
AlexIIL:
```suggestion
			Log.warn(LogCategory.SOLVING, String.format("Ignoring duplicate mod %s of the same version %s loaded from %s", id, version, from));
```
That would make it a noisy log error - I don't think `debug` shows by default.

src/main/java/org/quiltmc/loader/impl/plugin/QuiltPluginManagerImpl.java
AlexIIL:
```suggestion
		ModLoadOption current = set.byVersionSingles.get(version);
		if (current != null && current.isMandatory() && mod.isMandatory() && current.getClass() == mod.getClass() && QuiltLoader.isDevelopmentEnvironment()) {
```

AlexIIL (changes requested):
I'll admit, I don't like it. But as it's definitely easier to do this here than figure out what's going wrong in loom or multiproject builds in general I'd be happy to see this added.

I'd like to be able to *disable* this functionality with a system property in future though.

AlexIIL (approved):

PR #272, opened 2023-04-01
Switch hashes to SHA-1, and add them to the mod table.

AlexIIL:
This is intended for use by automatic log-parsing tools like [cozy](https://github.com/QuiltMC/cozy-discord) to check for more information about the mods present. In theory we could include dependency information in the table, but that's going to take a lot of space.

This doesn't add them to the plugin table quite yet - which needs to happen before this is merged.


AlexIIL:
Oh er I'm dumb and should have pinged earlier - @gdude2002 does this do what you need?

gdude2002:
This looks like it should be fine, yep

gdude2002 (approved):

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -732,6 +757,7 @@ public void appendModTable(Consumer<String> to) {
 			// - ID
 			// - version
 			// - loader plugin
+			// - SHA-512
```
TheGlitch76:
```suggestion
			// - SHA-1
```

TheGlitch76 (commented):
I'm not in love with this approach. I'm already cursing the mod table size any time I have to use mclo.gs , and this just makes it even worse! While this PR does allow modders to be sure that users are using a real, untampered JAR published from an official source, I'm not convinced that the value in such a usecase outweighs the increased table size. I'd probably rather have an unreadable, 1-line index->hash map dumped at the bottom of the table.

On the other hand, the table is already uncontrollably large, especially when things are super nested (which is inevitable in larger modpacks). An extra 25 characters isn't the end of the world, so I won't block this feature from being merged as-is.

Gaming32:
The primary usecase, though, is, from what I understand, looking up mod information on Modrinth to provide information for log parsing, as opposed to being used for file validation.

TheGlitch76:
Yes, i understand. A one-line map would be sufficient for the primary usecase.

AlexIIL (commented):
Blocked until #284 is merged (or refused).

PR #273, opened 2023-04-01
CRITICAL: Make Minecraft entrypoint patch support FlexVer

LemmaEOF:
Makes `EntrypointPatch` in the `minecraft` subdir use `Version.parse` instead of `SemanticVersion.parse`. It turns out Mojang cannot be trusted to use SemVer-compliant versioning, as [Snapshot 23w13a_or_b crashes Quilt immediately](https://mclo.gs/5aMCn9D)! This should just be a one-line fix.

### DIFF
--- a/minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/EntrypointPatch.java
+++ b/minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/EntrypointPatch.java
@@ -571,7 +571,7 @@ public class EntrypointPatch extends GamePatch {
 
 	private Version getGameVersion() {
 		try {
-			return SemanticVersion.parse(gameProvider.getNormalizedGameVersion());
+			return Version.parse(gameProvider.getNormalizedGameVersion());
 		} catch (VersionParsingException e) {
 			throw new RuntimeException(e);
 		}


sschr15 (approved):

LambdAurora (approved):

TheKodeToad:
Doesn't this have exactly the same behaviour? I'm just looking at the GitHub files but they seem to both end up calling SemanticVersionImpl.of

LemmaEOF:
The difference is that Version.of calls SemanticVersionImpl.of then *falls back to FlexVer* if that fails - going just through SemanticVersion means it can't do that fallback in cases like this

TheKodeToad:
> The difference is that Version.of calls SemanticVersionImpl.of then _falls back to FlexVer_ if that fails - going just through SemanticVersion means it can't do that fallback in cases like this

Completely missed that, oops!

TheKodeToad (approved):
LGTM!

violet-sh (approved):

williambl (approved):

sschr15:
For anyone looking for a release of this now for testing and such, I've built this under `org.quiltmc:quilt-loader:0.19.0-beta.3+local` in `https://maven.concern.i.ng/releases/`

LemmaEOF:
Effectively merged! Closing up now that it's redundant.

PR #275, opened 2023-04-02
Provide Fabric Loader 0.4.19

comp500:
Fabric Loader was updated to support 23w13a_or_b, so the version number was bumped (which FAPI depends on)

See https://github.com/FabricMC/fabric-loader/commit/69fefcb72af0598524b1468028b3a93202dc517b

### DIFF
--- a/src/main/resources/quilt.mod.json
+++ b/src/main/resources/quilt.mod.json
@@ -22,7 +22,7 @@
     "provides": [
       {
         "id": "fabricloader",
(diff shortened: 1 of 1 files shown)


TheGlitch76:
Merged as https://github.com/QuiltMC/quilt-loader/commit/21ca6e8766d5a43d47329e202b70e4f56d589317; our git structure has fixes go on `release` first and then merged into `develop`.

PR #276, opened 2023-04-02
Fix addFileToScan only adding files as non-direct

comp500:
Plugins should be required to specify whether files should be direct or not, so sub-mods and top-level mods can be handled properly. (breaking change to the experimental plugin API!)

Most plugins will want to add files as direct anyway; the javadoc even implied that this was the case (matching the mods dir). This also fixes mods added by argument being non-direct (so they wouldn't be loaded unless a dependency was added on them).

Includes fixes to the javadocs of ModLocation.isDirect and QuiltPluginContext.addFileToScan.


AlexIIL (approved):

TheGlitch76:
https://github.com/QuiltMC/quilt-loader/commit/999ed3018c76d686626dc9181dffb29008092420

PR #280, opened 2023-04-06
 Print a mod "type" (Quilt, Fabric) instead of its loading plugin in mod table

TheGlitch76:
Mod types are a lot clearer to a user than plugins. For debugging, a single-line map is printed at the bottom of the mod table which gives mappings from mod id to all of the mod kinds it has provided.

An image is worth several words (outdated screenshot):
An image of the mod table's appearance after this PR. The old "Plugin" column is replaced by a new "Kind" column

Two other changes were made to the table in this PR:
- The table now ignores case when sorting the table, putting "Minecraft" next to "minecraft_test". This makes the mod table easier to search, since capitalization is unpredictable in mod names.
- The mod table now adds a `Mod Table Version` property at the end, which is incremented any time we make a change to the table or the extra variables at the end. This is intended to give @gdude2002 a little help when parsing logs in Cozy.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -664,7 +664,7 @@ public final class QuiltLoaderImpl {
 		int maxNameLength = "Mod".length();
 		int maxIdLength = "ID".length();
 		int maxVersionLength = "Version".length();
-		int maxKindLength = "Kind".length();
+		int maxTypeLength = "Kind".length();
 		List<Integer> maxSourcePathLengths = new ArrayList<>();
 		Path absoluteGameDir = gameDir.toAbsolutePath().normalize();
 		Path absoluteModsDir = modsDir.toAbsolutePath().normalize();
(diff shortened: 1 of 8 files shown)


gdude2002 (approved):
Makes sense really - mod jar hash when tho?? :>

TheKodeToad (approved):
sorry if it's annoying reviewing as someone who isn't part of the org :P

Looks pretty good!
Non-critical thoughts:
- Hyphen-Seperated-Words isn't very readable.
- Type is possibly a clearer word.
- The plugin list could be another table in future?

AlexIIL (changes requested):
To be honest I agree with @TheKodeToad here - I don't think `Mod-Table-Version` or `Plugin-Kinds` should be hyphenated (nothing else that loader logs is hyphenated - with the exception of `Sub-Files`).

Other than that I'm happy with this. (I'm not sure how I feel about "kind" TBH, but it works).

AlexIIL (changes requested):
The output looks good!

I'm blocking this until #284 is merged (or is refused).

TheKodeToad (approved):
Looks better now!

src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
```diff
@@ -694,6 +694,16 @@ public void appendModTable(Consumer<String> to) {
...
 
+
+		HashMap<String, Set<String>> types = new HashMap<>();
+
+		for (ModContainerExt mod : mods) {
+			types.computeIfAbsent(mod.pluginId(), k -> new HashSet<>()).add(mod.modType());
+		}
+
+		table.addExtraData("Mod Table Version: 2");
+		table.addExtraData("Plugin Types: " + types);
```
AlexIIL: I don't really like the idea of "extra data" being added on to a table - especially since you can just emit those lines directly? (In the future we could turn the plugin types into a proper table too - which would be better off as its own table rather than appended to the existing table)

AlexIIL (approved):

PR #283, opened 2023-04-08
Only use one `Sub-File` column

TheGlitch76:
This drastically reduces the size of the mod table when mods have heavily nested JiJs, like in Universal Graves.
Jars are chained with an `!`, creating strings like the following:
`
/META-INF/jars/polymer-core-0.4.0-pre.2+1.19.4-rc2.jar!/META-INF/jars/polymer-networking-0.4.0-pre.2+1.19.4-rc2.jar!/META-INF/jars/polymer-common-0.4.0-pre.2+1.19.4-rc2.jar!/META-INF/jars/packet-tweaker-0.3.1+23w06a.jar
`

Comparison mod tables:
[before.txt](https://github.com/QuiltMC/quilt-loader/files/11184389/before.txt)
[after.txt](https://github.com/QuiltMC/quilt-loader/files/11184390/after.txt)

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -665,7 +665,9 @@ public final class QuiltLoaderImpl {
 		int maxIdLength = "ID".length();
 		int maxVersionLength = "Version".length();
 		int maxPluginLength = "Plugin".length();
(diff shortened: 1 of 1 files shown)


AlexIIL (changes requested):
This looks good!

I'm only blocking this until #284 is merged or declined, since it significantly simplifies table generation (and merging PRs like these)

AlexIIL (approved):

Issue #285, opened 2023-04-09
QuiltBaseFileSystem doesn't support 'glob' path matching.


AlexIIL:
https://github.com/QuiltMC/quilt-loader/blob/2a45297b4e95966d11c90e2d1e2ece1f6794daff/src/main/java/org/quiltmc/loader/impl/filesystem/QuiltBaseFileSystem.java#L211

TheGlitch76:
https://github.com/QuiltMC/quilt-loader/commit/54e48387ba39f064b52bc2ebfc87361d93d97113

PR #286, opened 2023-04-09
Copy google jimfs GlobToRegex class to implement glob pattern matching

AlexIIL:
I think I've followed the same license changes as the json5 project did - although this definitely needs checking!

Implements #285.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/filesystem/GlobToRegex.java
+++ b/src/main/java/org/quiltmc/loader/impl/filesystem/GlobToRegex.java
@@ -1,0 +1,486 @@
+/*
+ * Copyright 2023 QuiltMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *		 http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


src/main/java/org/quiltmc/loader/impl/filesystem/GlobToRegex.java
```diff
@@ -0,0 +1,486 @@
...
+ * - Repackaged.
+ * - "InternalCharMatcher" was partially copied into this file
+ * - "com.google.common.base.Preconditions.checkNotNull" was replaced with "java.lang.Objects.requireNonNull"
+ *
+ * You may view the original, including its license header, here:
+ * https://github.com/google/jimfs/blob/64b0b187221a529be531271d9e27627ad1954572/jimfs/src/main/java/com/google/common/jimfs/GlobToRegex.java
+ * 
+ * In addition the file "InternalCharMatcher" was partially copied into this file, under the same license, from here:
+ * https://github.com/google/jimfs/blob/4739aee7e8e6ee9c66801541921f42f9af93387d/jimfs/src/main/java/com/google/common/jimfs/InternalCharMatcher.java
+ */
```
TheGlitch76, line 31: Yeah, this is fine. It's technically way overboard for what the license requires, but it's (1) best not to poke the bear and (2) useful for when someone else comes down the line in 7 years (which isn't super unrealistic, considering parts of Loader still present today were written in 2016) and has to figure out wtf this is.

TheGlitch76 (approved):

PR #289, opened 2023-04-10
Add mod vetoes

### LINKED ISSUE #287: Mods with load_type ALWAYS but environment CLIENT crash the solver on a dedicated server
[Crash report example](https://github.com/QuiltMC/quilt-loader/files/11186327/crash-2023-04-09_22.51.19.2257-quilt_loader.txt)

This occurs because `OptionalModIdDefinition` handles a load type of ALWAYS by adding a rule that the mod must be present, but we disable mods in a server environment by adding a rule requiring the mod is never loaded with DisabledModIdDefinition.

TheGlitch76:
Summary of what a mod veto is, directly from the javadoc:
> Mod vetoes allow plugins to remove load options from the mod list immediately before it is finalized and passed to the transform cache, after solving. The returned options are removed with no checks of any kind; they simply disappear from the mod list. Since the solver will not be re-ran, no errors will occur due to missing dependencies or the like.
> Vetoes are performed in one pass: every plugin sees the initial mod list and submits vetoes, which are collected and applied together. Then, all plugins are made aware of vetoed mods (including the ones they vetoed themselves) through QuiltLoaderPlugin.onModVeto.
> **This is intended to allow plugins to "disable" mods that shouldn't load in the current environment, but need to be present during solving.**

This is a solution to #287 that exactly replicates the behavior of old Quilt Loader/Fabric Loader, but without special-casing the mechanism. In the future, we can move the removal of mods based on environment to a specific Minecraft Loader Plugin with almost no extra effort.

This PR targets the `release` branch (0.18) because the feature is directly related to a bugfix, and the bugfix is needed for QSL (and I'd like to get QSL on "modern" Loader as soon as possible)

### DIFF
--- a/minecraft/minecraft-test/src/main/resources/quilt.mod.json
+++ b/minecraft/minecraft-test/src/main/resources/quilt.mod.json
@@ -9,10 +9,14 @@
       "main": [
         "net.fabricmc.minecraft.test.FabricEntrypointTest"
       ]
-    }
+    },
+    "load_type": "always"
   },
   "mixin": [
     "fabricmc.test.mixins.client.json"
   ],
+  "minecraft": {
+    "environment": "client"
+  },
   "experimental_chasm_transformers_disabled": "chasm_transformers"
 }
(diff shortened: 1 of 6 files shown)


AlexIIL:
I'm not sure about adding vetoes to be honest - I'd rather information was available to the solver if possible. In addition this is quite a large change to go on a release branch, so I've made #290 as a less intrusive change.

PR #305, opened 2023-05-08
Improved Error Messages

TheGlitch76:
This is **far from complete**, but I'm opening now so that other people can contribute if they like.
I'd like to get this merged in the `0.20` or `0.21` cycles.

Instead of looking at mandatory mods and walking down the chain to find errors, this PR looks at errors and walks up the chain to determine sources. This lets us collect data and display it in a clearer and denser manner.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.1.1-all.zip
 zipStoreBase=GRADLE_USER_HOME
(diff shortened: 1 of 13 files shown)


AlexIIL:
Here's a before-and-after comparison of putting QSL for 1.18.2 into the mods folder (when you've got minecraft 1.20 pre 1).

0.19.0-beta.13:
Image 1

This PR:
Image 2

AlexIIL:
Compared to the previous error messages I think this is *worse* in some ways, but *better* in others:

It's **better** because:

- It all fits in a single screen! Previously there were **54** error messages, but now there's only 4.
- It gets rid of the "transitively" stuff and just shows the mod that depends on something which is missing - which makes it much easier to debug at a glance.

It's *worse* because:

- 3 of the 4 errors are misleading - they are about mods being "incompatible versions" despite the *actual* error being minecraft.
- You removed the icons, and I liked those :p

Overall I'm not sure the approach of "finding errors" makes sense - the whole "graph" of rules collectively describe a single error, so it doesn't make sense to ignore some parts of it in favor of others.

PR #311, opened 2023-05-13
Deprecate `quilt-json5`, update Quilt Config

TheGlitch76:
This PR is not to be merged until the beginning of the `0.20` cycle.

This PR brings the following changes:
- Deprecate the quilt-json5 exposed to mods. We shouldn't expose libraries directly in Loader unless we absolutely must (in case we want to break them!), and there's no reason to expose quilt-json5 here. It is being replaced by QUP (name to be changed at a later date), and mods are responsible for including it themselves if they want it.
- Update Quilt Config to `1.1.0`.
  - This includes the new `ReflectiveConfig` interface, a replacement to `WrappedConfig` that is actually useful (you can modify the config now! yay!)
  - Before `0.20` is stable, we'll need to work out all the kinks in this new interface, and anything else we might need to change in Quilt Config.

I'll need to fix up the javadocs on the v2 `QuiltConfig` interface before merge.


TheKodeToad:
qulit :iea:

PR #312, opened 2023-05-14
Alternative improvements to the solver

AlexIIL:
This uses a few of the general improvements that @TheGlitch76 wrote in #305, but doesn't discard the previous solver helper - which results in this error screen when adding the wrong version of QSL into the mods folder (compare against [these images](https://github.com/QuiltMC/quilt-loader/pull/305#issuecomment-1546785194))

Screenshot from 2023-05-14 02-32-12

Similar to #305 this has a few problems: some mods are included multiple times in the list (likely due to their dependency chains being slightly different), and it doesn't show what those dependency chains are. I'm not sure how important this is though.

(This also fixes "provided" mods not being recognized by the solver helper)

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/plugin/SolverErrorHelper.java
+++ b/src/main/java/org/quiltmc/loader/impl/plugin/SolverErrorHelper.java
@@ -498,18 +498,14 @@ class SolverErrorHelper {
 
 	static class UnhandledError extends SolverError {
 
-		final List<Collection<Rule>> errors = new ArrayList<>();
+		final Collection<Rule> rules;
 
 		public UnhandledError(Collection<Rule> rules) {
-			errors.add(rules);
+			this.rules = rules;
 		}
 
 		@Override
 		boolean mergeInto(SolverError into) {
-			if (into instanceof UnhandledError) {
-				((UnhandledError) into).errors.addAll(errors);
-				return true;
-			}
 			return false;
 		}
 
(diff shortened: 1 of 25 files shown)


AlexIIL:
As of https://github.com/QuiltMC/quilt-loader/pull/312/commits/4bbd18c08467457559ccd6a2537a13ca62be3557 this also shows the full list of rules when we encounter an "unhandled" solver error, and renames it to a "complex" error:
Screenshot from 2023-05-25 16-53-56

src/main/java/org/quiltmc/loader/impl/plugin/SolverErrorHelper.java
```diff
@@ -517,4 +482,145 @@ private static boolean reportDuplicateMandatoryMods(QuiltPluginManagerImpl manag
...
+		}
+
+		@Override
+		void report(QuiltPluginManagerImpl manager) {
+			QuiltDisplayedError error = manager.theQuiltPluginContext.reportError(
+				QuiltLoaderText.translate("error.unhandled_solver")
+			);
+			error.appendDescription(QuiltLoaderText.of("error.unhandled_solver.desc"));
+			error.addOpenLinkButton(QuiltLoaderText.of("button.quilt_loader_report"), "https://github.com/QuiltMC/quilt-loader/issues");
+			error.appendReportText("Unhandled solver error involving the following rules:");
```
TheGlitch76: Since there are still some super common cases that are unhandled (like `breaks`), this should probably go to the forum, not our issue tracker, since it's more of a support question than a "please fix this" question 99 times out of 100.

AlexIIL: Yes, good point.

AlexIIL: Okay, fixed.

TheGlitch76 (approved):
we're definitely not done with this code yet but this is 100% an improvement

PR #329, opened 2023-07-03
Add Loader support for `quilt.mod.json5`

EnnuiL:
Creating this PR as a draft until support on Quilt Loom's side is worked on, but so far? I believe the Loader-side support is solid! Reviews are appreciated

The benefits of JSON5 will be great for developers


EnnuiL:
Work on Quilt Loom has been pushed as [quilt-loom#36](https://github.com/QuiltMC/quilt-loom/pull/36)!

src/main/java/org/quiltmc/loader/impl/plugin/quilt/StandardQuiltPlugin.java
```diff
@@ -242,12 +242,44 @@ private ModLoadOption[] scan0(Path root, QuiltLoaderIcon fileIcon, ModLocation l
 		PluginGuiTreeNode guiNode) throws IOException {
 
 		Path qmj = root.resolve("quilt.mod.json");
-		if (!FasterFiles.isRegularFile(qmj)) {
+		Path qmj5 = root.resolve("quilt.mod.json5");
+		Path usedQmj = qmj;
+
+		if (FasterFiles.exists(qmj5)) {
+			boolean enableQmj5 = QuiltLoader.isDevelopmentEnvironment() || Boolean.getBoolean(SystemProperties.ENABLE_QUILT_MOD_JSON5);
+			if (!QuiltLoader.isDevelopmentEnvironment() && !enableQmj5) {
```
AlexIIL: I don't think you should be able to enable this in production, at all - since it allows mods to require the flag at runtime.

EnnuiL: Ah, I can remove the property; It was more like, an emergency thing than something people should use, but hm, you do have a point; `-Dloader.development=true` already exists after all

EnnuiL:
As a note, I've noticed that Quilt Parsers hasn't been relocated properly; I went ahead and fixed that

EnnuiL:
I went ahead and expanded the QMJ5 handling by having the Loader error out if a QMJ is also found, which prevents from further deviation from its intended purpose;
The funny thing was that with that, I managed to catch an issue where mods with duplicate metadata were being iterated through twice! And this extended to mods containing both a QMJ and a FMJ! I do wonder if it could be done in a more effective way though;
anyway, here's the screenshot!
A Quilt Loader error message that says "A coexistence of 'quilt.mod.json' and 'quilt.mod.json5' has been detected. These metadata files cannot coexist with each other due to their conflicting purposes!"

EnnuiL:
In order to prevent trouble (although with some trouble being caused as a side-effect), I've made the QMJ5 support opt-in, so the build system has to explicitly ask for it in order for things to work;
This may make migrating from a QMJ workspace to a QMJ5 one more difficult though, so I'll have to tackle that issue myself somehow
A Quilt Loader error message explaining that it can't read 'quilt.mod.json5' files without its support being enabled

EnnuiL:
Update: by using the Fabric DLI system, a migration from an already-existing QMJ workspace to a QMJ5 one should be smooth

EnnuiL:
> Update: by using the Fabric DLI system, a migration from an already-existing QMJ workspace to a QMJ5 one should be smooth

Tackled by [quilt-loom#37](https://github.com/QuiltMC/quilt-loom/pull/37)

EnnuiL:
oh boy, here we go;
Rebased to QLoader 0.26.4, see [quilt-loom#53](https://github.com/QuiltMC/quilt-loom/pull/53) for new companion PR

TheGlitch76 (approved):

AlexIIL:
I'm sorry for taking this long - and thanks for updating this!

PR #330, opened 2023-07-09
Update dependencies

EnnuiL:
This has the important side-effect of Letting Me Use Java 20 On Development! :p
Also, FlexVer had enjoyed few improvements that we should pull in


AlexIIL:
Sorry for the delay - this looks good, thanks!

PR #335, opened 2023-07-16
fix a typo in `InternalsHiderTransform`

eristhea:
world's most nitpicky quilter forks loader just to fix one typo


AlexIIL:
Thanks!

PR #352, opened 2023-08-24
[0.21] Refactor TransformCache, apply QuiltTransformer etc. statically

TheGlitch76:
This PR does not include non-mod class support, since it is currently unimplemented anyway.


src/main/java/org/quiltmc/loader/impl/transformer/ChasmInvoker.java
```diff
@@ -241,9 +238,12 @@ private static <T> T this_value_is_actually_nullable(T in) {
 	static class QuiltMetadata {
 
 		final ModLoadOption from;
+		// the dot name of this class
+		final String name;
```
AlexIIL, line 242: Should this be Javadoc? Eclipse doesn't show non-javadoc comments above fields as javadoc.

src/main/java/org/quiltmc/loader/impl/discovery/ClasspathModCandidateFinder.java
```diff
@@ -104,6 +70,37 @@ public static void findCandidatesStatic(ModAdder out) {
...
+		Enumeration<URL> quiltMods = QuiltLauncherBase.getLauncher().getTargetClassLoader().getResources(metadataName);
+		while (quiltMods.hasMoreElements()) {
+			URL url = quiltMods.nextElement();
+
+			try {
+				Path path = LoaderUtil.normalizeExistingPath(UrlUtil.getCodeSource(url, metadataName));
+				List<Path> paths = pathGroups.get(path);
+
+				if (paths == null) {
+                    if (!url.getProtocol().equals("jar")) {
```
AlexIIL, line 83: There's a few places like this were you've used spaces instead of tabs - did you setup your environment again? (This isn't actually a problem though, just something we'll replace at some point I guess?)

AlexIIL (approved):

AlexIIL:
This PR should be merged before #356, since it's bigger :p

PR #356, opened 2023-08-26
Fix mod hash computation

AlexIIL:
This should fix the transform cache being recomputed every launch due to system mods not being skipped.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/util/FileHasherImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/util/FileHasherImpl.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2022, 2023 QuiltMC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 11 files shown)


src/main/java/org/quiltmc/loader/impl/util/FileHasherImpl.java
```diff
@@ -0,0 +1,131 @@
...
+				}
+			});
+		} catch (UncheckedIOException e) {
+			throw e.getCause();
+		}
+
+		return Arrays.copyOf(hash, HASH_LENGTH);
+	}
+
+	private static byte[] computeRecursiveHash0(Path path) throws IOException {
```
TheGlitch76, line 104: Question--will this report as a "SHA" hash in the mod table?

AlexIIL: Good question! No, actually - the mod table doesn't use `ModLoadOption.computeOriginHash` at all, as it instead uses `computeNormalHash` for every source path.

TheGlitch76 (approved):

TheGlitch76:
Merged manually with https://github.com/QuiltMC/quilt-loader/commit/94e5ca747bdad4de32bb39cec09a02926f994610 as you've disabled my ability to push directly to your branch.

PR #369, opened 2023-09-25
Treat "miscGameLibraries" as part of Minecraft for the transform cache

### LINKED ISSUE #357: Transform cache does not include Minecraft's dependencies
Minecraft dependencies visible to mixin need to be present in the transform cache too. I would recommend shading them all into the Minecraft mod load option.

TheGlitch76:
Closes #357


AlexIIL:
This is good - or at least it's better than *not* doing this, but in the future I'd like to look into exposing libraries as mods somehow, with their own modids? That will also help in cases where libraries have a common file, as they won't get shared properly.

PR #371, opened 2023-10-03
Remap AW into `getTargetNamespace` instead of named

heipiao233:
Since [L69](https://github.com/QuiltMC/quilt-loader/blob/develop/src/main/java/org/quiltmc/loader/impl/transformer/RuntimeModRemapper.java#L69) is `getTargetNamespace`, [L135](https://github.com/PillowMC/quilt-loader/blob/develop/src/main/java/org/quiltmc/loader/impl/transformer/RuntimeModRemapper.java#L135) should also be `getTargetNamespace`.


AlexIIL (approved):

AlexIIL:
Yes, thanks!

PR #374, opened 2023-10-09
Add solver pre-processing

AlexIIL:
For some reason the way we use sat4j seems to be non-optimal, leading to the full 5 seconds of optimising being used. (Removing this limit in the blanketcon pack leads to Sat4j churning for at least 20 mins to "prove" that it's found the most optimal result, with this PR Sat4j only takes ~60ms to solve the [remaining rules and options](https://gist.github.com/AlexIIL/86ba6059b4c6fbbd2b7a151667bf2608)).

This means we can be more confident that we get a fully optimal solution in more cases. (When this isn't the case you'll see this line in the log: `Aborted mod solving optimisation due to timeout`)

This also means we can use the solver system for more features, and be confident that it still finds the optimal solution quickly.

Breaking changes:

- Plugins can no longer override `LoadOption.equals` and `LoadOption.hashCode`

Misc Changes:

- Sat4jWrapper can now have rules changed at anytime, and correctly reverts back to solve it properly.

Other changes from the src/main/resources/changelog/solver-pre-processor.txt file:

* Added a pre-processor step to mod solving.
    * This greatly reduces the time taken for Sat4j to optimise the final mod set (pick the newest possible versions of every mod)
        * Previously this was capped at 5 seconds.
    * Added a warning when picking between two mods when we don't have any reason to pick one over the other.
        * This can happen when two mods provide each other, or two mods provide the same version of a third mod.
    * Added a system property to disable this "loader.mod_solving.disable_pre_processor"
    * Added a system property to explain what this pre-processor is struggling with (partially simplified mod sets)
        * "loader.mod_solving.print_results" prints unsolved sub-problems, and the final chosen options.
        * This is an alternative to the very verbose "loader.debug.mod_solving" property.
    * Made the log line "Aborted mod solving optimisation due to timeout" always print when it happens.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ group = org.quiltmc
 description = The mod loading component of Quilt
 url = https://github.com/quiltmc/quilt-loader
 # Don't forget to change this in QuiltLoaderImpl as well
-quilt_loader = 0.21.0-temp.29
+quilt_loader = 0.21.0-temp.33
 
 # Fabric & Quilt Libraries
 asm = 9.5
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ group = org.quiltmc
 description = The mod loading component of Quilt
 url = https://github.com/quiltmc/quilt-loader
 # Don't forget to change this in QuiltLoaderImpl as well
-quilt_loader = 0.21.0-temp.38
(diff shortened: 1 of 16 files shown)


AlexIIL:
This will probably go in 0.22, since it's a bit bigger than the other PRs that are going in 0.21.2

PR #378, opened 2023-10-18
Add support for 1.20.3 snapshots

LostLuma:
Fixes current snapshots still being read as `1.20.2-alpha.xx.xx.x`.


AlexIIL:
Looks good, thanks!

PR #380, opened 2023-10-23
Add ModEntrypoint as an API for AdapterLoadableClassEntry

AlexIIL:
Since the only public changes are generics, existing mods which use `getEntrypoints()` (which they shouldn't as it's all either `PLUGIN_API` or `LEGACY_EXPOSED`) will continue to work in binary (compiled) form, but not source form.

`ModEntrypoint` intentionally doesn't expose the `adapter` or `value` at the moment since only loader itself should need it, at least in theory.

This also removes the `plugins` field from the QMJ reader, since it used to use `AdapterLoadableClassEntry` but isn't actually hooked up anywhere (as actual plugins don't get defined in this way)

This also has two extra API changes:

- Adds `ModLicense.fromIdentifier` & `fromIdentifierOrDefault`, which exposes the same methods in `ModLicenseImpl`
- Adds `QuiltPluginManager.loadZipNow`, which is an immediate version of `loadZip`

### DIFF
--- a/src/main/java/org/quiltmc/loader/api/ModLicense.java
+++ b/src/main/java/org/quiltmc/loader/api/ModLicense.java
@@ -18,9 +18,13 @@
 package org.quiltmc.loader.api;
 
 import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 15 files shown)


AlexIIL:
This will go in 0.22, since it's a bit more breaking than other changes.

PR #395, opened 2024-01-09
Support sided mixin configs

OroArmor:


src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
```diff
@@ -1008,4 +1009,30 @@ public static VersionRange readVersionSpecifier(String string) throws VersionFor
...
+				if (config == null) {
+					throw parseException((JsonLoaderValue) value, "Mixin entry inside must have a config value");
+				}
+				LoaderValue environment = object.get("environment");
+
+				if (environment == null) {
+					destination.computeIfAbsent(EnvType.CLIENT, (env) -> new ArrayList<>()).add(config.asString());
+					destination.computeIfAbsent(EnvType.SERVER, (env) -> new ArrayList<>()).add(config.asString());
+				} else {
+					destination.computeIfAbsent(EnvType.valueOf(environment.asString().toUpperCase()), (env) -> new ArrayList<>()).add(config.asString());
```
AlexIIL: Two problems:

- `Enum.valueOf` throws an exception if the passed name isn't a valid enum constant, but this should use `parseException()` or use an alternative test (a custom switch?)
- This should accept only the same names as the existing environment field: https://github.com/QuiltMC/rfcs/blob/main/specification/0002-quilt.mod.json.md#the-environment-field.

src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
```diff
@@ -1008,4 +1009,30 @@ public static VersionRange readVersionSpecifier(String string) throws VersionFor
...
+				destination.computeIfAbsent(EnvType.CLIENT, (env) -> new ArrayList<>()).add(value.asString());
+				destination.computeIfAbsent(EnvType.SERVER, (env) -> new ArrayList<>()).add(value.asString());
+			} else if (value.type() == LoaderValue.LType.OBJECT) {
+				LoaderValue.LObject object = value.asObject();
+				LoaderValue config = object.get("config");
+
+				if (config == null) {
+					throw parseException((JsonLoaderValue) value, "Mixin entry inside must have a config value");
+				}
+				LoaderValue environment = object.get("environment");
```
AlexIIL: You need to test if `environment` is a `LoaderValue.LType.STRING` or not before converting it to one. (Since `LoaderValue.asString()` throws a `ClassCastException` if it's not a string, which doesn't contain any information about what part of the json isn't a string).

src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
```diff
@@ -1008,4 +1009,30 @@ public static VersionRange readVersionSpecifier(String string) throws VersionFor
...
 	}
+
+	private static void readMixins(JsonLoaderValue.ArrayImpl array, Map<EnvType, List<String>> destination) {
+		for (LoaderValue value : array) {
+			if (value.type() == LoaderValue.LType.STRING) {
+				destination.computeIfAbsent(EnvType.CLIENT, (env) -> new ArrayList<>()).add(value.asString());
+				destination.computeIfAbsent(EnvType.SERVER, (env) -> new ArrayList<>()).add(value.asString());
+			} else if (value.type() == LoaderValue.LType.OBJECT) {
+				LoaderValue.LObject object = value.asObject();
+				LoaderValue config = object.get("config");
```
AlexIIL: You need to test if `environment` is a `LoaderValue.LType.STRING` or not before converting it to one. (Since `LoaderValue.asString()` throws a `ClassCastException` if it's not a string, which doesn't contain any information about what part of the json isn't a string)

AlexIIL (changes requested):
This needs to be an [RFC](https://github.com/QuiltMC/rfcs/blob/main/specification/0002-quilt.mod.json.md#the-environment-field) before we can merge this unfortunately. (I've considered writing an RFC to move the spec to this repo instead, but for now this change needs to be there too).

OroArmor:
Rfc has been merged!

AlexIIL (approved):

PR #407, opened 2024-02-16
Fix multiple quilt config classes being on the classpath, leading to some classes being missing.

AlexIIL:
This also fixes loader including quilt config (in it's entirety) and the jetbrain/intelij annotations.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,7 +73,9 @@ sourceSets {
 }
 
 configurations {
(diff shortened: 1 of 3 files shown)


eristhea (approved):
I knew I shouldn't have updated those launchwrapper files :p

TheGlitch76 (approved):

PR #410, opened 2024-03-12
Bootstrap tests

AlexIIL:
Uses https://github.com/QuiltMC/quilt-loader-bootstrap to allow loader to not be on the root classpath. (Although the bootstrap must be on the root classpath instead).

This will make it easier for users to update loader without the installer, and for a proper "server bootstrap" to be written.

~~This depends on https://github.com/QuiltMC/quilt-parsers/pull/4, and as such it doesn't build yet.~~


TheGlitch76:
Parsers 0.3.0 will be out in a few minutes which includes https://github.com/QuiltMC/quilt-parsers/pull/4

PR #413, opened 2024-03-19
Print the break reason on error

EnnuiL:
Oh god I can't believe that `reason` was useless all this time;
This PR makes `reason` useful for `break`s:
(Note: "Reason" has been changed to "Breaking mod's reason")
image

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/plugin/SolverErrorHelper.java
+++ b/src/main/java/org/quiltmc/loader/impl/plugin/SolverErrorHelper.java
@@ -810,7 +810,7 @@ class SolverErrorHelper {
 			}
 
 			for (ModLoadOption mod : from) {
-				error.appendReportText("- " + manager.describePath(mod.from()), "");
+				error.appendReportText("- " + manager.describePath(mod.from()));
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


UpcraftLP (approved):
LGTM 👍

src/main/java/org/quiltmc/loader/impl/plugin/SolverErrorHelper.java
```diff
@@ -797,8 +805,12 @@ void report(QuiltPluginManagerImpl manager) {
 			report.append(", which is present!");
 			error.appendReportText(report.toString(), "");
 
+			if (!reason.isEmpty()) {
+				error.appendReportText("Breaking mod's reason: " + reason, "");
+			}
+
 			for (ModLoadOption mod : from) {
-				error.appendReportText("- " + manager.describePath(mod.from()));
+				error.appendReportText("- " + manager.describePath(mod.from()), "");
```
AlexIIL: Was this additional newline intentional? I'm not sure it's necessary here.
AFAIK it would make the report look like this:
```
BadCraft-A breaks all versions of libnetworkstack, which is present!

Breaking mod's reason: I don't like LibNetworkStack lol

- <mods>/badcraft-1.0.0.jar!/META-INF/jars/badcraft-submod-a.jar

- <mods>/badcraft-1.0.0.jar!/META-INF/jars/badcraft-submod-b.jar

- <mods>/badcraft-1.0.0.jar!/META-INF/jars/badcraft-submod-c.jar

- <mods>/badcraft-1.0.0.jar!/META-INF/jars/badcraft-submod-d.jar
```

AlexIIL (approved):
Thanks for hooking this up!

src/main/java/org/quiltmc/loader/impl/plugin/SolverErrorHelper.java
EnnuiL: Oh no it was not

PR #414, opened 2024-03-20
Feature: `@Requires` Annotation

FirstMegaGame4:
This pull request adds a new annotation => `@Requires` which allows to put some strings in it specifying required mods for the targeted code to not be stripped by the loader.

The best example I have is the use of Trinkets API when you want to create an item that would optionally use the `Trinket` interface when the Trinkets mod is detected:

`MyOptionalTrinketItem.java`
```java
public class MyOptionalTrinketItem extends Item implements @Requires("trinkets") Trinket {

	public MyOptionalTrinketItem(Settings settings) {
    		super(settings);
	}

	@Override
	@Requires("trinkets")
	public Multimap<EntityAttribute, EntityAttributeModifier> getModifiers(ItemStack stack, SlotReference slot, LivingEntity entity, UUID uuid) {
		// ...
	}
}
```


src/main/java/org/quiltmc/loader/impl/transformer/QuiltTransformer.java
```diff
@@ -34,15 +35,17 @@
 
 import net.fabricmc.accesswidener.AccessWidenerClassVisitor;
 
+import org.quiltmc.loader.impl.util.StrippingDataContainer;
+
 @QuiltLoaderInternal(QuiltLoaderInternalType.NEW_INTERNAL)
 final class QuiltTransformer {
-	public static byte @Nullable [] transform(boolean isDevelopment, EnvType envType, TransformCache cache, AccessWidener accessWidener, String name, ModLoadOption mod, byte[] bytes) {
+	public static byte @Nullable [] transform(boolean isDevelopment, EnvType envType, TransformCache cache, AccessWidener accessWidener, String name, ModLoadOption mod, List<ModLoadOption> modList, byte[] bytes) {
```
AlexIIL: `TransformCache` exposes the mod list in `getMods()`, so this additional parameter unnecessary.

src/main/java/org/quiltmc/loader/api/minecraft/Requires.java
```diff
@@ -0,0 +1,58 @@
...
+ * of this annotation, you could use it in a new package-info.java file of the current package dedicated for the
+ * mod integration, in order to trim the complete package, which will provide an additional safety layer by avoiding
+ * errors when loading and / or calling classes and elements that should not when the required mod is not present. */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.TYPE, ElementType.TYPE_USE, ElementType.FIELD, ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.PACKAGE })
+@Documented
+public @interface Requires {
+
+	/** @return required mods */
+	String[] mods();
```
AlexIIL: If you name this `value` rather than `mods` then it can be omitted when used - for example:

```java
@Requires("buildcraftenergy")
public class CompatWithBuildCraftEnergy
```
rather than
```java
@Requires(mods = "buildcraftenergy")
public class CompatWithBuildCraftEnergy
```

src/main/java/org/quiltmc/loader/impl/launch/knot/KnotClassDelegate.java
```diff
@@ -322,9 +326,17 @@ boolean computeCanLoadPackage(String pkgName, boolean allowFromParent) {
...
-			return !data.stripEntirePackage;
+			PackageStrippingDataContainer data = new PackageStrippingDataContainer();
+
+			ClassReader reader = new ClassReader(bytes);
+
+			PackageEnvironmentStrippingVisitor envStripping = new PackageEnvironmentStrippingVisitor(QuiltLoaderImpl.ASM_VERSION, data, envType);
+			reader.accept(envStripping, ClassReader.SKIP_CODE | ClassReader.SKIP_FRAMES);
+
+			PackageRequiresStrippingVisitor requiresStripping = new PackageRequiresStrippingVisitor(QuiltLoaderImpl.ASM_VERSION, data, modCodeSourceMap);
+			reader.accept(requiresStripping, ClassReader.SKIP_CODE | ClassReader.SKIP_FRAMES);
```
AlexIIL: I can understand this separation (it's cleaner name-wise at least) but please combine `@Requires` annotation processing with environment annotation processing - that way we don't need to traverse the class files *twice*. (This should also greatly reduce the number of changes you need to make in this PR).

src/main/java/org/quiltmc/loader/impl/transformer/EnvironmentStrippingVisitor.java
```diff
@@ -36,9 +33,11 @@
 import net.fabricmc.api.EnvironmentInterface;
 import net.fabricmc.api.EnvironmentInterfaces;
 
+import org.quiltmc.loader.impl.util.StrippingDataContainer;
+
 /** Scans a class for Environment and EnvironmentInterface annotations to figure out what needs to be stripped. */
 @QuiltLoaderInternal(QuiltLoaderInternalType.LEGACY_EXPOSED)
-public class EnvironmentStrippingData extends ClassVisitor {
+public class EnvironmentStrippingVisitor extends ClassVisitor {
```
AlexIIL: Don't rename classes - especially those which are `LEGACY_EXPOSED` as mods can access them. I have no idea *why* mods might access/use this, but I'd prefer to *not* break mods unless absolutely necessary.

AlexIIL (changes requested):
I like `@Requires`! It's not a feature that's needed often, but it's still better to put this properly in loader than anywhere else. (It's also a better name than what I was thinking of originally -`@ModOnly` - when I added `@ClientOnly`).

As for the actual PR changes, overall I'd like to see a *reducation* in the number of changes made - please put the `@Requires` processing into `PackageEnvironmentStrippingData` and `EnvironmentStrippingData` so you don't need to duplicate the code in a few places. (Sorry about the bad names - they are exposed to mods so I'd rather not break them unless necessary).

src/main/java/org/quiltmc/loader/impl/transformer/EnvironmentStrippingData.java
```diff
@@ -16,265 +16,142 @@
...
-	private static final String ENVIRONMENT_INTERFACES_DESCRIPTOR = Type.getDescriptor(EnvironmentInterfaces.class);
+	private static final String ENVIRONMENT_DESCRIPTOR = StrippingData.ENVIRONMENT_DESCRIPTOR;
+	private static final String ENVIRONMENT_INTERFACE_DESCRIPTOR = StrippingData.ENVIRONMENT_INTERFACE_DESCRIPTOR;
+	private static final String ENVIRONMENT_INTERFACES_DESCRIPTOR = StrippingData.ENVIRONMENT_INTERFACES_DESCRIPTOR;
 
 	// Quilt annotations
-	private static final String CLIENT_ONLY_DESCRIPTOR = Type.getDescriptor(ClientOnly.class);
-	private static final String SERVER_ONLY_DESCRIPTOR = Type.getDescriptor(DedicatedServerOnly.class);
+	private static final String CLIENT_ONLY_DESCRIPTOR = StrippingData.CLIENT_ONLY_DESCRIPTOR;
+	private static final String SERVER_ONLY_DESCRIPTOR = StrippingData.SERVER_ONLY_DESCRIPTOR;
```
AlexIIL: You don't need to keep any of the *private* fields or classes since external mods can't access them.

FirstMegaGame4: Just thought that they could be accessed via accessors or somewhat, but now that you're telling me that I don't think mixin can be applied to Quilt Loader classes...

src/main/java/org/quiltmc/loader/impl/transformer/EnvironmentStrippingData.java
```diff
@@ -16,265 +16,134 @@
...
-	private final Collection<String> stripInterfaces = new HashSet<>();
-	private final Collection<String> stripFields = new HashSet<>();
-	private final Collection<String> stripMethods = new HashSet<>();
-
-	/** Every method contained in this will also be contained in {@link #stripMethods}. */
-	private final Collection<String> stripMethodLambdas = new HashSet<>();
-
-	private class FabricEnvironmentAnnotationVisitor extends AnnotationVisitor {
-		private final Runnable onEnvMismatch;
+	private class FabricEnvironmentAnnotationVisitor extends StrippingData.FabricEnvironmentAnnotationVisitor {
```
AlexIIL: You missed a class! This is private too :p

src/main/java/org/quiltmc/loader/impl/transformer/EnvironmentStrippingData.java
```diff
@@ -16,265 +16,134 @@
...
+			super.visitEnum(name, descriptor, value);
 		}
 	}
 
-	private class QuiltEnvironmentAnnotationVisitor extends AnnotationVisitor {
-		private final Runnable onEnvMismatch;
-		private final Runnable onEnvMismatchLambdas;
-
-		private boolean stripLambdas = true;
+	private class FabricEnvironmentInterfaceAnnotationVisitor extends StrippingData.FabricEnvironmentInterfaceAnnotationVisitor {
```
AlexIIL: (Also private - please delete)

src/main/java/org/quiltmc/loader/impl/transformer/EnvironmentStrippingData.java
```diff
@@ -16,265 +16,134 @@
...
-				onEnvMismatchLambdas.run();
-			}
+			super.visitEnd();
 		}
 	}
 
-	private class FabricEnvironmentInterfaceAnnotationVisitor extends AnnotationVisitor {
-		private boolean envMismatch;
-		private Type itf;
+	private class QuiltEnvironmentAnnotationVisitor extends StrippingData.QuiltEnvironmentAnnotationVisitor {
```
AlexIIL: Here's one

src/main/java/org/quiltmc/loader/impl/transformer/EnvironmentStrippingData.java
```diff
@@ -16,265 +16,134 @@
...
 		@Override
 		public void visitEnd() {
-			if (envMismatch) {
-				stripInterfaces.add(itf.getInternalName());
-			}
+			super.visitEnd();
 		}
 	}
 
-	private AnnotationVisitor visitMemberAnnotation(String descriptor, boolean visible, Runnable onEnvMismatch,
```
AlexIIL, line 137: Wow, there's a lot

AlexIIL (changes requested):

src/main/java/org/quiltmc/loader/api/Requires.java
```diff
@@ -0,0 +1,58 @@
...
+
+package org.quiltmc.loader.api;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/** Applied to declare that the annotated element specific mods to exist.
```
TheGlitch76: I think you a word.
```suggestion
/** Applied to declare that the annotated element requires specific mods to exist.
```

FirstMegaGame4: Oh, it seems I not concentrated.

AlexIIL (approved):

AlexIIL:
Thanks! I'm going to tweak some minor things locally though - mostly just `List<String> mods` -> `Set<String> mods`  as that should be an easy optimization? Otherwise this looks good!

PR #418, opened 2024-04-01
Publish dependencies as seperate module

lukebemish:
Sets up publishing of loader's dependencies in a loom-friendly fashion. As I had to touch a few different parts of the buildscript while doing this, I took the occasion to switch to using toolchains to handle the java version used for compiling and generating javadoc.

The dependencies are publishes as a module at `org.quiltmc:quilt-loader-dependencies`; currently, loader itself does not publish a dependency on this module, though this would be changed once an accompanying exclusion has been added to quilt loom, backported to old versions, and some time given for people to update

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,8 +18,13 @@ plugins {
 	id 'com.github.johnrengelman.shadow' version '8.1.1'
 }
 
-sourceCompatibility = JavaVersion.VERSION_1_8
-targetCompatibility = JavaVersion.VERSION_1_8
+java {
+	toolchain {
+		languageVersion.set(JavaLanguageVersion.of(8))
+	}
+	withJavadocJar()
+	withSourcesJar()
+}
 
 archivesBaseName = "quilt-loader"
 version = project.quilt_loader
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -18,8 +18,13 @@ plugins {
 	id 'com.github.johnrengelman.shadow' version '8.1.1'
 }
 
-sourceCompatibility = JavaVersion.VERSION_1_8
```
TheGlitch76, line 21: I believe vscode/eclipse still needs these lines to work?

OroArmor (commented):
Should we move to a version manifest now?

build.gradle
lukebemish: Does someone have eclipse to test this? Toolchains are at this point not a new gradle feature; I'd be rather concerned if eclipse didn't support them.

lukebemish:
What do you mean by a version manifest? Like using a version catalog, or something in terms of publishing?

OroArmor:
yeah, the `libs.versions.toml`

lukebemish:
That's not a bad idea but that would be a different PR, I think

AlexIIL (approved):

build.gradle
lukebemish: This was addressed on discord: it works fine, luckily.

PR #421, opened 2024-04-04
Changed Icons class to be public

coolGi69:
Currently, there is no way to set the [V1ModMetadataBuilder](https://github.com/QuiltMC/quilt-loader/blob/56a4df3a0dc624965596a6003e277230ee0f78bc/src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataBuilder.java#L57)'s icon due to the Icons class being private. This PR just turns it public so that it can be used by external mods who need a way to set their icons.


AlexIIL:
Thanks!

PR #427, opened 2024-05-01
Fix Fabric mod author and contributor potentially duplicating mod contributors

LostLuma:
Fix edge case when a Fabric mod has the same name in both the `authors` and `contributors` list producing two separate mod contributors during translation instead of a single one with two roles.

I'm not sure if there is actually any mods with this sort of duplicate metadata out there in the wild but I ran into it when testing something earlier and figured it'd be good to fix it anyway for correctness sake.


TheGlitch76 (approved):
Simple enough, good catch!

PR #428, opened 2024-05-04
Allowed the GameProvider to provide the namespace

coolGi69:
This PR adds a new optional function to the GameProvider called `getNamespace` to allow it to provide the namespace of the current session (which should be either be "named" or "intermediary").\
It also exposes the namespace to the GamePatch allowing it to be passed as an argument to it.

~~This PR also removes [these lines in TinyFDPatch](https://github.com/QuiltMC/quilt-loader/blob/5e36a6f0d615bf3d07261ed94c26850d6910280e/minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/TinyFDPatch.java#L57) as (to the best of my knowledge), it will always return false (`launcher.getMappingConfiguration().getTargetNamespace()` and `QuiltLoader.getMappingResolver().getNamespaces()` are always the same), unless something went wrong (where the namespace is called something other than "named" or "intermediary").~~
Reverted this change as it will be needed for the planned `hashed` mappings/namespace (https://github.com/QuiltMC/quilt-loader/pull/428#discussion_r1590394769)

### DIFF
--- a/minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/TinyFDPatch.java
+++ b/minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/TinyFDPatch.java
@@ -51,8 +51,15 @@ public final class TinyFDPatch extends GamePatch {
 			return;
 		}
 
+		String className = MORE_OPTIONS_DIALOG_CLASS_NAME;
 
-		final ClassNode classNode = context.getClassNode(MORE_OPTIONS_DIALOG_CLASS_NAME);
+		// Only remap the classname when needed to prevent loading the mappings when not required in prod.
+		if (!launcher.getMappingConfiguration().getTargetNamespace().equals("intermediary")
+				&& QuiltLoader.getMappingResolver().getNamespaces().contains("intermediary")) {
+			className = QuiltLoader.getMappingResolver().mapClassName("intermediary", MORE_OPTIONS_DIALOG_CLASS_NAME);
+		}
+
+		final ClassNode classNode = context.getClassNode(className);
 
 		if (classNode == null) {
 			// Class is not present in this version, nothing to do.
--- a/minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/TinyFDPatch.java
+++ b/minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/TinyFDPatch.java
@@ -45,21 +45,14 @@ public final class TinyFDPatch extends GamePatch {
 	private static final String DIALOG_TITLE = "Select settings file (.json)";
 
 	@Override
-	public void process(QuiltLauncher launcher, GamePatchContext context) {
+	public void process(QuiltLauncher launcher, String namespace, GamePatchContext context) {
 		if (launcher.getEnvironmentType() != EnvType.CLIENT) {
 			// Fix should only be applied to clients.
 			return;
 		}
 
-		String className = MORE_OPTIONS_DIALOG_CLASS_NAME;
 
-		// Only remap the classname when needed to prevent loading the mappings when not required in prod.
-		if (!launcher.getMappingConfiguration().getTargetNamespace().equals("intermediary")
-				&& QuiltLoader.getMappingResolver().getNamespaces().contains("intermediary")) {
-			className = QuiltLoader.getMappingResolver().mapClassName("intermediary", MORE_OPTIONS_DIALOG_CLASS_NAME);
-		}
-
-		final ClassNode classNode = context.getClassNode(className);
+		final ClassNode classNode = context.getClassNode(MORE_OPTIONS_DIALOG_CLASS_NAME);
 
 		if (classNode == null) {
 			// Class is not present in this version, nothing to do.
--- a/src/main/java/org/quiltmc/loader/impl/entrypoint/GamePatch.java
+++ b/src/main/java/org/quiltmc/loader/impl/entrypoint/GamePatch.java
@@ -128,12 +128,20 @@ public abstract class GamePatch {
 		return ((access & 0x0F) == (Opcodes.ACC_PUBLIC | 0 /* non-static */));
 	}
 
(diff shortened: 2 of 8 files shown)


src/main/java/org/quiltmc/loader/impl/entrypoint/GamePatch.java
```diff
@@ -128,12 +129,18 @@ protected boolean isPublicInstance(int access) {
 		return ((access & 0x0F) == (Opcodes.ACC_PUBLIC | 0 /* non-static */));
 	}
 
-	public void process(QuiltLauncher launcher, Function<String, ClassReader> classSource, Consumer<ClassNode> classEmitter) {
+	public void process(QuiltLauncher launcher, String namespace, Function<String, ClassReader> classSource, Consumer<ClassNode> classEmitter) {
 		throw new AbstractMethodError(getClass() + " must override one of the 'process' methods!");
 	}
+	public void process(QuiltLauncher launcher, Function<String, ClassReader> classSource, Consumer<ClassNode> classEmitter) {
+		process(launcher, QuiltLoaderImpl.INSTANCE.tryGetGameProvider().getNamespace(), classSource, classEmitter);
```
AlexIIL: `tryGetGameProvider()` should be replaced with `getGameProvider()` (it has a better error message if it's not set).

src/main/java/org/quiltmc/loader/impl/entrypoint/GamePatch.java
```diff
@@ -128,12 +129,18 @@ protected boolean isPublicInstance(int access) {
...
+	}
 
+	public void process(QuiltLauncher launcher, String namespace, GamePatchContext context) {
+		process(launcher, namespace, context::getClassSourceReader, context::addPatchedClass);
+	}
 	public void process(QuiltLauncher launcher, GamePatchContext context) {
-		process(launcher, context::getClassSourceReader, context::addPatchedClass);
+		process(launcher, QuiltLoaderImpl.INSTANCE.tryGetGameProvider().getNamespace(), context);
```
AlexIIL: `tryGetGameProvider()` should be replaced with `getGameProvider()` (it has a better error message if it's not set).

src/main/java/org/quiltmc/loader/impl/launch/common/MappingConfiguration.java
```diff
@@ -78,7 +80,7 @@ public TinyTree getMappings() {
 	}
 
 	public String getTargetNamespace() {
-		return QuiltLauncherBase.getLauncher().isDevelopment() ? "named" : "intermediary";
+		return QuiltLoaderImpl.INSTANCE.tryGetGameProvider().getNamespace();
```
AlexIIL: `tryGetGameProvider()` should be replaced with `getGameProvider()` (it has a better error message if it's not set).

minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/TinyFDPatch.java
```diff
@@ -45,21 +45,14 @@ public final class TinyFDPatch extends GamePatch {
...
-		String className = MORE_OPTIONS_DIALOG_CLASS_NAME;
 
-		// Only remap the classname when needed to prevent loading the mappings when not required in prod.
-		if (!launcher.getMappingConfiguration().getTargetNamespace().equals("intermediary")
-				&& QuiltLoader.getMappingResolver().getNamespaces().contains("intermediary")) {
-			className = QuiltLoader.getMappingResolver().mapClassName("intermediary", MORE_OPTIONS_DIALOG_CLASS_NAME);
-		}
-
-		final ClassNode classNode = context.getClassNode(className);
+		final ClassNode classNode = context.getClassNode(MORE_OPTIONS_DIALOG_CLASS_NAME);
```
AlexIIL: Revert these changes. `QuiltLoader.getMappingResolver().getNamespaces()` returns a set, which will contain just  `[ "intermediary" ]` at runtime and `[ "named", "intermediary" ]` in a development environment. (In the future this could also contain `hashed` if/when we implement it in loader).

AlexIIL (changes requested):
I agree with most of these changes - but with 1 disagreement, and 4 nitpicks. (`tryGetGameProvider()` should only be used over `getGameProvider()` if you can do something with a null return value, and it's invoked early enough to be null, but since you just use it directly there's no reason not to use `getGameProvider()` instead)

src/main/java/org/quiltmc/loader/impl/entrypoint/GamePatch.java
coolGi69: I have found a better way to do it that uses the [MappingConfiguration](https://github.com/QuiltMC/quilt-loader/pull/428/files#diff-0d54a473dc8b9723b082ad8a8a27309ee719e13c863cc58580ad0aaf479cf025) instead, and can get the namespace from the provided launcher varuable.

minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/patch/TinyFDPatch.java
coolGi69: Ah, I forgot that `hashed` mappings were planned. I've re-added the bit of code now, so it should be fine :+1:

src/main/java/org/quiltmc/loader/impl/launch/common/MappingConfiguration.java
coolGi69: There is a chance that this gets called before the game provider is set, so I just changed it to check if it got a valid game provider or not.

coolGi69:
@AlexIIL not sure if I should have notified you around 2 days ago when I made the commit, but those suggestions have been implemented in the most recent commit.\
Hopefully that should be all, but do tell me if there is anything else that needs to be changed.

Edit: Turns out there is a request for re-review button on GH that I missed, sry about that.

TheGlitch76 (approved):

AlexIIL (approved):

PR #430, opened 2024-05-12
Fix knot class loading allowlist not working

### LINKED ISSUE #251: Certain legacy server versions crash while trying to load mods

LostLuma:
Resolves #251.

The method of looking up which prefixes were (dis)allowed was flawed as it always looked up using the entire path to a class file (including the part in the jar), whereas the allowlist only maps from the path to the jar to allowed prefixes in the jar.

I've also changed `UrlUtil.asUrl` to normalize the path as otherwise it produces URLs with a full stop at the current working directory, which caused issues with looking up the prefixes as well. See an equivalent example:
```diff
- Path.of("./src/hello.txt").toUri().toURL();
- $1 ==> file:/C:/Users/lilly/projects/quilt-loader/./src/hello.txt
+ Path.of("./src/hello.txt").toAbsolutePath().normalize().toUri().toURL();
+ $2 ==> file:/C:/Users/lilly/projects/quilt-loader/src/hello.txt
```

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/launch/knot/KnotClassDelegate.java
+++ b/src/main/java/org/quiltmc/loader/impl/launch/knot/KnotClassDelegate.java
@@ -244,11 +244,19 @@ class KnotClassDelegate {
 			}
 		}
 
-		if (!allowedPrefixes.isEmpty()) {
+		if (!allowedPrefixes.isEmpty() && url != null) {
+			String fileName = LoaderUtil.getClassFileName(name);
+			URL codeSource = null;
+
+			try {
+				codeSource = UrlUtil.getSource(fileName, url);
+			} catch (UrlConversionException e) {
+				Log.warn(LogCategory.GENERAL, "Failed to get the code source URL for " + url);
+			}
+
 			String[] prefixes;
 
-			if (url != null
-					&& (prefixes = allowedPrefixes.get(url.toString())) != null) {
+			if (codeSource != null && (prefixes = allowedPrefixes.get(codeSource.toString())) != null) {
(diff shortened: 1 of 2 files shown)


TheGlitch76 (approved):

AlexIIL (approved):
Sorry for the delay - this all looks good!

PR #432, opened 2024-05-20
Port to mapping-io (Glitch's Version)

TheGlitch76:
This is step 0 in adding the ability for Loader to use alternative intermediates for mods, or use something like mojmaps at runtime. This PR migrates the pre-existing MappingConfiguration to use mapping-io, but doesn't do much to actually refactor Loader to take advantage of mapping-io's strengths.

This PR moves a few dependencies to be shadowed that Floader moved as well. Tiny Remapper was required to be moved to work with mapping-io, tiny mapping parser was removed by upstream, and I just decided to bring AccessWidener in line with upstream as well.

(PS: I know QMT exists-ish, but it's not finished and I want to work on a modloader, not a mappings parser. When and if QMT is finished, we can evaluate migrating to it, but the potential to use a first-party library later shouldn't prevent us from improving Loader today)


src/main/java/org/quiltmc/loader/api/gui/QuiltLoaderIcon.java
```diff
@@ -32,7 +32,7 @@ public enum SubIconPosition {
 
 	QuiltLoaderIcon getDecoration(SubIconPosition position);
 
-	/** Returns a new icon with a sub-icon added to this main icon. The given sub-icon must not already have sub-icons.
+	/** Returns a new icon with a sub-icon added to this main icon. The given sub-ienihztvrcd con must not already have sub-icons.
```
OroArmor: what

eristhea: what

Pyrofab: what

FirstMegaGame4: dude stop talking in hashed

src/main/java/org/quiltmc/loader/impl/transformer/RuntimeModRemapper.java
```diff
@@ -95,16 +92,35 @@ public static void remap(TransformCache cache) {
...
 
-				remapper.readInputsAsync(tag, info.inputPath);
+				Path in = fs.getPath(mod.id());
+				// HACK: Tiny Remapper eagerly opens ZIP files contained in mods (i.e. the JAR files they've attempted to JiJ)
+				// This causes LOTS of problems involving duplicate classes (and potentially the wrong version of the class being selected!!!),
+				// so we ONLY expose the .class files to Tiny Remapper
+				Files.walk(mod.resourceRoot()).filter(p -> p.getFileName().toString().endsWith(".class")).forEach(p -> {
+					try {
+						Files.createDirectories(in.resolve(p.getParent().toAbsolutePath().toString()));
+						fs.mount(p, in.resolve(p.toAbsolutePath().toString()));
```
AlexIIL: Adding a `MountOption.READ_ONLY` will avoid a partial FS copy duing `fs.switchToReadOnly` (`fs.mount(p, in.resolve(p.toAbsolutePath().toString()), MountOption.READ_ONLY);`)

AlexIIL (approved):
This looks good now that I've pushed three small fixes!

I'm not that happy with removing/hiding libraries (as this breaks a few older mods) but I'll look into if we can fix them later, as I don't want that to prevent this from being merged.

PR #434, opened 2024-05-28
feat(loader): cli parameter to ignore unknown mods

### LINKED ISSUE #433: Skip unknown mods windows
Is there a way to skip the windown shown when there're files that are not mods? The `.dll` file just needs to be there :smile: 

image

voidpointer0x00:
Fixes #433 restoring old behavior _(not shore which version this GUI was introduced in)_ with extra cli argument `--ignore-unknown-mods`

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -42,6 +42,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Random;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


AlexIIL (changes requested):
Thanks for the PR! I agree that this is needed, but not quite with how this is implemented.

Specifically:
1. Quilt-loader generally uses system properties to configure its behavior - these are the static constants in [`SystemProperties`](https://github.com/QuiltMC/quilt-loader/blob/develop/src/main/java/org/quiltmc/loader/impl/util/SystemProperties.java). Can you replace the argument with a system property like `loader.ignore_unsupported_mods`? (This also means we don't pass it to minecraft, which logs a message like `Completely ignored arguments: [--ignore-unknown-mods]`)
2. I think this is the wrong place to filter out ignored mods - the window which gets hidden *could* contain unrelated warnings (in the `Files` tab) which should still be shown if the config option `always_show_mod_state_window` is enabled. Instead, this should probably go in `QuiltPluginManagerImpl.populateModsGuiTab` (where the warnings are generated) - basically just skip the unsupported mod checking if the system property is set to true.

voidpointer0x00:
@AlexIIL, thanks for the review, will redo :)

voidpointer0x00:
@AlexIIL, reimplemented, tested — that also works

AlexIIL (approved):

AlexIIL:
Thanks!

PR #435, opened 2024-05-30
Allowed the (knot) launcher to give the intended namespace

coolGi69:
Adds a `getFinalNamespace()` to the `QuiltLauncher` allowing the knot to provide other namespaces outside of `intermediary` to be used.
Currently, the access widener's namespace preference is hardcoded as `intermediary` within the `RuntimeModRemapper`, but with this change the value is now gotten from the launcher.

This PR also makes several functions and classes for the knot class loaders public, allowing for the creation of other knots.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ group = org.quiltmc
 description = The mod loading component of Quilt
 url = https://github.com/quiltmc/quilt-loader
 # Don't forget to change this in QuiltLoaderImpl as well
-quilt_loader = 0.28.0-beta.3
+quilt_loader = 0.28.0-beta.4
 
 # Fabric & Quilt Libraries
 asm = 9.7.1
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -131,7 +131,7 @@ public final class QuiltLoaderImpl {
 
 	public static final int ASM_VERSION = Opcodes.ASM9;
 
-	public static final String VERSION = "0.28.0-beta.3";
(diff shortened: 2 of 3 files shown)


coolGi69:
~~I think a good idea is to move the namespace stuff that I previously moved into the game provider to the launcher. (Less duplicated code)\
I'll implement that tomorrow, so don't merge this yet.~~

src/main/java/org/quiltmc/loader/impl/launch/knot/KnotClassDelegate.java
heipiao233, line 1: Why making this public?

770grappenmaker: > allowing for the creation of other knots

coolGi69:
Nvm, I don't see what past me in https://github.com/QuiltMC/quilt-loader/pull/435#issuecomment-2139377429 wanted to change.\
Sorry for the long open time, I was on a bit of a development hiatus.

I'll double-check this commit has no issues and report back on its status in a few minutes.

coolGi69:
Should be ready to be reviewed/merged now.

Note: I'm not sure if you'd prefer to revert https://github.com/QuiltMC/quilt-loader/pull/435/commits/89b42d3a2bb4ea2fde9f8f5ae9e90784d7606f44, but it will make it so that other knots can just build off the base knot, and don't have to re-implement everything.\
If it is accepted, then many of the things that are made public here can be turned back to private as they no longer have to be implemented by other peoples knots.

coolGi69:
For now, I'm closing this PR. When I have time over the next week, I'll update it to work with the latest commit of develop, and reopen this PR.

PR #446, opened 2024-08-08
Begin QMJ Writer

OroArmor:
This will allow people to make tools to convert fabric projects into quilt projects without needing to re-implement the whole QMJ spec

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
+++ b/src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
@@ -18,7 +18,6 @@ package org.quiltmc.loader.impl.metadata.qmj;
 
 import java.io.IOException;
 import java.io.Writer;
-import java.lang.reflect.Field;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -28,7 +27,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
-import java.util.TreeMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -41,8 +39,10 @@ import org.quiltmc.loader.api.ModDependency;
 import org.quiltmc.loader.api.ModLicense;
 import org.quiltmc.loader.api.ModMetadata;
 import org.quiltmc.loader.api.VersionConstraint;
+import org.quiltmc.loader.api.VersionInterval;
 import org.quiltmc.loader.api.VersionRange;
 import org.quiltmc.loader.api.plugin.ModMetadataExt;
+import org.quiltmc.loader.impl.fabric.metadata.V1ModMetadataFabric;
 import org.quiltmc.loader.impl.metadata.FabricLoaderModMetadata;
 import org.quiltmc.loader.impl.util.QuiltLoaderInternal;
 import org.quiltmc.loader.impl.util.QuiltLoaderInternalType;
@@ -52,8 +52,6 @@ import net.fabricmc.loader.api.metadata.ModEnvironment;
 
 @QuiltLoaderInternal(QuiltLoaderInternalType.NEW_INTERNAL)
 public class InternalModMetadataJsonWriter {
-	private static final List<String> SPECIFIC_KEY_ORDER = Stream.of("schema_version", "quilt_loader", "mixin", "access_widener", "minecraft").collect(Collectors.toList());
-
 	public static void write(InternalModMetadata mod, Writer writer) throws IOException {
 		JsonWriter json = JsonWriter.json(writer);
 		((JsonLoaderValue) toJson(mod)).write(json);
@@ -65,21 +63,7 @@ public class InternalModMetadataJsonWriter {
 	}
 
 	public static LoaderValue toJson(InternalModMetadata mod) {
-		Map<String, LoaderValue> values = new TreeMap<>((o1, o2) -> {
-			if (o1.equals(o2)) {
-				return 0;
-			}
-
-			for (String key : SPECIFIC_KEY_ORDER) {
-				if (o1.equals(key)) {
-					return -1;
-				} else if (o2.equals(key)) {
-					return 1;
-				}
-			}
-
-			return o1.compareTo(o2);
-		});
+		Map<String, LoaderValue> values = new LinkedHashMap<>();
 
 		values.put("schema_version", new JsonLoaderValue.NumberImpl("", 1));
 		values.put("quilt_loader", createQuiltLoader(mod));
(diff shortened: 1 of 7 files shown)


src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
```diff
@@ -0,0 +1,435 @@
...
+					if (icon == null) {
+						return null;
+					}
+
+					return string(icon);
+				} else if (icons instanceof Icons.Multiple) {
+					Field mapField = Icons.Multiple.class.getDeclaredField("icons");
+					mapField.setAccessible(true);
+					SortedMap<Integer, String> map = (SortedMap<Integer, String>) mapField.get(icons);
+					return object(obj -> map.forEach((size, icon) -> obj.put(Integer.toString(size), string(icon))));
```
AlexIIL: You should change these fields to package-private, rather than using reflection.

src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
```diff
@@ -0,0 +1,435 @@
...
+			for (String key : SPECIFIC_KEY_ORDER) {
+				if (o1.equals(key)) {
+					return -1;
+				} else if (o2.equals(key)) {
+					return 1;
+				}
+			}
+
+			return o1.compareTo(o2);
+		});
```
AlexIIL: Just use a `LinkedHashMap` here - it will have the same effect, without needing to write a custom comparator.

src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
```diff
@@ -0,0 +1,435 @@
...
+						return string((String) stringField.get(icons));
+					} else if (icons.getClass().getName().contains("MapEntry")) {
+						Field iconsField = icons.getClass().getDeclaredField("icons");
+						iconsField.setAccessible(true);
+						SortedMap<Integer, String> map = (SortedMap<Integer, String>) iconsField.get(icons);
+						return object(obj -> map.forEach((size, icon) -> obj.put(Integer.toString(size), string(icon))));
+					}
+				} catch (Exception e) {
+					throw new RuntimeException(e);
+				}
```
AlexIIL, line 302: You should change these fields to public rather than using reflection.

(We could debate about if we *should* expose these, considering those objects are `LEGACY_EXPOSED`, but there's no cleaner way to expose this inside the same project, and it's unlikely for quilt mods to depend on being able to access *fabric* mod internals).

src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
```diff
@@ -0,0 +1,435 @@
...
+		throw new IllegalStateException("Unknown Mod Dependency type! " + dep.getClass().getName());
+	}
+
+	private static JsonLoaderValue version(VersionRange range) {
+		// TODO: test this
+		if (range.equals(VersionRange.ANY)) {
+			return string("*");
+		}
+
+		return array(range.convertToConstraints().stream().map(VersionConstraint::toString).map(InternalModMetadataJsonWriter::string));
```
AlexIIL: It shouldn't be possible to convert a version range to a single array of version constraints - so I think `VersionRangeImpl.convertToConstraints()` is actually bugged.

This should also serialize using the new `object` format rather than arrays - as arrays will cause quilt loader to emit a deprecation warning. The resultant object would look like this for a VersionRange composed of 2 VersionIntervals:

```json
{
  "any": [ { "all": [ *array-1* ] }, { "all": [ *array-2* ] } ]
}
```

Using something like this code:

```java
LoaderValue writeRange(VersionRange range) {
  if (range.size() == 1) {
    VersionInterval interval = range.get(0);
    if (interval.equals(VersionInterval.ALL) {
      return string("*");
    }
    return writeInterval(interval);
  }

  List<LoaderValue> sub = new ArrayList<>();
  for (VersionInterval interval : range) {
    sub.add(writeInterval(interval));
  }

  Map<String, LoaderValue> map = new LinkedHashMap<>();
  map.put("any", array(sub));
  return object(map);
}

LoaderValue writeInterval(VersionInterval interval) {
  VersionRange createdRange = VersionRange.of(interval);
  // Convert to constraints will be accurate if there's only a single interval
  List<LoaderValue> out = new ArrayList<>();
  for (VersionConstraint constraint : createdRange.convertToConstraints()) {
    out.add(string(constraint.toString()));
  }

  if (out.size() == 1) {
    return out.get(0);
  } else {
    Map<String, LoaderValue> sub = new LinkedHashMap<>();
    sub.put("all", array(out));
    return object(sub);
  }
}
```

src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
```diff
@@ -0,0 +1,435 @@
...
+
+	private static JsonLoaderValue.ObjectImpl object(Map<String, LoaderValue> values) {
+		return new JsonLoaderValue.ObjectImpl("", values);
+	}
+
+	private static JsonLoaderValue.ObjectImpl object(Consumer<Map<String, LoaderValue>> consumer) {
+		Map<String, LoaderValue> values = new LinkedHashMap<>();
+		consumer.accept(values);
+		return object(values);
+	}
```
AlexIIL, line 424: In theory it could be cleaner to use `org.quiltmc.loader.api.plugin.LoaderValueFactory` here, but the rest of the codebase doesn't :/

(Oh, and `JsonLoaderFactoryImpl.object` uses a `HashMap` rather than a `LinkedHashMap` so that would need to be changed before it could be used)

AlexIIL (changes requested):
I'm somewhat conflicted about this - I agree that this functionality is useful (and needed), but unhappy about adding apis to internal packages. I can understand that there isn't really a better place for this though - it's only intended to be used by external tools that will already have to use internal code like `org.quiltmc.loader.impl.metadata.FabricModMetadataReader` and some related classes.

Overall the code is reasonable, with a couple of major problems:

- Reflection shouldn't be used to access specific fields or classes inside the same project - especially if access can be made package-private. Alternately the `getIcon` method could be made optional methods in `InternalModMetadata`, `Icons`, and `FabricLoaderModMetadata`.
- Version ranges as arrays are deprecated by the QMJ spec, in preference of explicit all/any specifiers. I've included some code which hopefully fixes this?

src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
OroArmor: Didn't break after making it compile!

src/main/java/org/quiltmc/loader/impl/metadata/qmj/InternalModMetadataJsonWriter.java
OroArmor: I migrated what i could into the methods to make it use less reimplementation, but there was some stuff I couldn't change without larger changes to loader value factory.

OroArmor:
> I'm somewhat conflicted about this - I agree that this functionality is useful (and needed), but unhappy about adding apis to internal packages. I can understand that there isn't really a better place for this though - it's only intended to be used by external tools that will already have to use internal code like `org.quiltmc.loader.impl.metadata.FabricModMetadataReader` and some related classes.

Yeah that's the problem. It both needs to be "API" but also not. It *probably* wont get run when loader is actually running, its just something that a tool can use when compiled against loader.

AlexIIL (approved):

Issue #217, opened 2023-01-22
FAPI crashes on beta 58 and above


SplendidAlakey:
**MC** 1.19.2
**FAPI** 0.73.0

While most Quilt users will/should use QSL, some will not and instead rely on JiJ'd modules the mods supply, e.g. Sodium and Iris combo.

On beta 58 and above having any FAPI modules results in `Failed to locate Minecraft assets root!` crash.

Logs: https://gist.github.com/SplendidAlakey/f44549d5a5a08825de9e54380f97538d

TheGlitch76:
This issue does not seem to affect 1.19.3

SplendidAlakey:
Fixed in beta 74.

PR #453, opened 2024-10-12
Optimize boot times by not copying to JAR when unnecessary

EnnuiL:
Once upon a time, a commit happened where in order to fix shenanigans, a deal with the devil was done: if you had Fabric Resource Loader API v0 installed, no matter if you had the Quilted variant, you were going to suffer long load times. This was a terrible deal! But every deal has a fix..

This PR optimizes up to ten seconds (or more! I haven't measured properly) of load times by simply not copying the jar unnecessarily every time the game starts, making dev envs snappier and player happier (I am happier lmfao)

This has been tested on Minecraft 1.21.1, 1.20.1, 1.19.4, 1.19.3, 1.19.2 and 1.18.2 in order to determine the bounds as well as whenever QFAPI was a sure way of fixing the crash

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/QuiltLoaderImpl.java
@@ -485,7 +485,7 @@ public final class QuiltLoaderImpl {
 				// Versions before 1.17.1 don't work very well if they aren't at the root of their zip file
 				return true;
 			}
-			if (modIds.contains("fabric-resource-loader-v0")) {
+			if (modIds.contains("fabric-resource-loader-v0") && !modIds.contains("quilted_fabric_resource_loader_v0")) {
 				if (Version.of("1.19.3").compareTo(mod.version()) > 0) {
 					// Fabric API turns minecraft into a resource pack to load from instead of using the classpath,
 					// so it also doesn't work very well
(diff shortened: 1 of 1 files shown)


EnnuiL:
Actually? Here's a decent before/after from my dev env:
Before:
The timings log from Quilt Loader 0.26.4, showing 7 seconds for "temp jar copy"
After:
The timings log from the pull request's Quilt Loader, showing 0 seconds for "temp jar copy"
It was that bad!

OroArmor (approved):
Looks great! I'm not super sure how this system works so some more documentation on what it's used for would be nice.

EnnuiL:
> Looks great! I'm not super sure how this system works so some more documentation on what it's used for would be nice.

As far as I know, it's used to fix issues like #217, where this is meant to fix expectations about how the filesystem works; it does have the cost though of uh, being slow. I'm not Alex though, I'm just bothered by how slow QLoader was augymlwulamgulym

AlexIIL:
Thanks for looking at this - I'd like to look into optimizing `tmp jar copy` if possible, since 7 seconds seems very slow for a single jar?

Either way this change seems good!

AlexIIL:
...Oh, one thing that did confuse me is that the `modIds` passed to `shouldCopyToJar` are only meant to contain "direct" mods, and *not* provided mods - did QFAPI change how this is provided?

EnnuiL:
> ...Oh, one thing that did confuse me is that the `modIds` passed to `shouldCopyToJar` are only meant to contain "direct" mods, and _not_ provided mods - did QFAPI change how this is provided?

Nope, the QFAPI provides were kept intact ever since I quiltified QFAPI

EnnuiL:
!!! It turns out you are right, if something else provides `fabric-resource-loader-v0`, it won't pass the check; I don't know how I missed that during the tests, but yeah, I guess this fix is more for a QFAPI-less era now

(this means that the additional check can be removed, with the only important one being the MC version check)

PR #464, opened 2025-01-01
 Allowed The (Knot) Launcher To Give The Intended Namespace 

coolGi69:
Continuation of https://github.com/QuiltMC/quilt-loader/pull/435. Now moved under the CRModders namespace.

<br/>

Adds a `getFinalNamespace()` to the `QuiltLauncher` allowing the knot to provide other namespaces outside of `intermediary` to be used.
Currently, the access widener's namespace preference is hardcoded as `intermediary` within the `RuntimeModRemapper`, but with this change the value is now gotten from the launcher.

This PR also makes several functions and classes for the knot class loaders public, allowing for the creation of other knots.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/discovery/ClasspathModCandidateFinder.java
+++ b/src/main/java/org/quiltmc/loader/impl/discovery/ClasspathModCandidateFinder.java
@@ -173,7 +173,7 @@ public class ClasspathModCandidateFinder {
 
 	public static Path getGameProviderPath() {
 		try {
-			return UrlUtil.asPath(QuiltLoaderImpl.INSTANCE.getGameProvider().getClass().getProtectionDomain().getCodeSource().getLocation());
+			return UrlUtil.asPath(QuiltLoaderImpl.class.getProtectionDomain().getCodeSource().getLocation());
 		} catch (Throwable t) {
 			Log.debug(LogCategory.DISCOVERY, "Could not retrieve launcher code source!", t);
 			return null;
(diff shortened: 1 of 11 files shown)


src/main/java/org/quiltmc/loader/impl/discovery/ClasspathModCandidateFinder.java
```diff
@@ -173,7 +173,7 @@ public static Path getLoaderPath() {
 
 	public static Path getGameProviderPath() {
 		try {
-			return UrlUtil.asPath(QuiltLoaderImpl.INSTANCE.getGameProvider().getClass().getProtectionDomain().getCodeSource().getLocation());
+			return UrlUtil.asPath(QuiltLoaderImpl.class.getProtectionDomain().getCodeSource().getLocation());
```
AlexIIL: This change looks a bit weird - this method is intended to find the game provider jar to load plugins from it - does that functionality not work for you? (Or do you set the launcher so `getLoaderPath` actually returns the game provider path by mistake?)

(This method is used next to `getLoaderPath()`, so it should probably not return the same location)

coolGi69:
This PR is no longer needed since the release of Quilt 0.29's refactor (thanks!).\
Although, commits https://github.com/QuiltMC/quilt-loader/pull/464/commits/5453bd5924aa8ecf954548702b23d03cf69d9297 https://github.com/QuiltMC/quilt-loader/pull/464/commits/d9450fff304325934406957560cecc51dd4f044a https://github.com/QuiltMC/quilt-loader/pull/464/commits/565c7f26b1b440d3f8acade2a6e6d6f20d7ee168 should probably still be added so external Knots be created.

PR #466, opened 2025-01-07
Make Classes Public For Extra Loaders Through Plugins

coolGi69:
This PR just makes some classes public to allow for loader plugins to be created.\
While both classes that are now made public aren't *necessary* to be to make a plugin, it makes it slightly easier.

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/gui/PluginIconImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/gui/PluginIconImpl.java
@@ -44,12 +44,12 @@ public final class PluginIconImpl extends QuiltGuiSyncBase implements QuiltLoade
 	final IconType icon;
 	final IconType[] subIcons = new IconType[4];
 
-	PluginIconImpl() {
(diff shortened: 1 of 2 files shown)


AlexIIL:
You should be able to make plugin icons via `QuiltLoaderGui.createIcon(*)`? (You're explicitly not meant to create `String path` icons yourself since they only read from loader's jar file).

Similarly, would you be okay if we added public api factory methods for `ModDependencyImpl` (in `ModDependency`) instead of making the classes public?

coolGi69:
> You should be able to make plugin icons via `QuiltLoaderGui.createIcon(*)`? (You're explicitly not meant to create `String path` icons yourself since they only read from loader's jar file).

Ah, alright. Missed that part when looking through the code when attempting to make the plugin.

> Similarly, would you be okay if we added public api factory methods for `ModDependencyImpl` (in `ModDependency`) instead of making the classes public?

Yeah, sure. That's fine, I was just thinking to reduce code duplication this could be public.\
I guess it's fine to implement it myself.

PR #469, opened 2025-01-25
Add experimental support for using Mojmap as a runtime

EnnuiL:
~~This expands on @TheGlitch76's initial efforts on running Mojmap as runtime; It works! The log messages are all translated to Mojmap instead of Intermediary! And in theory? Mods that only have code without any use of Mixins and AWs can work!~~

~~The issue is: the remapper is wonky! It currently hates Mixins and AWs *and* some results may be partial! So my experiment is y'all's problem now :p~~

~~(PS: The removal of Tiny v1 support will require Quilt Meta to provide Intermediary v2, which I have tested and can confirm it'll be fine)~~

^ Outdated -- see https://github.com/QuiltMC/quilt-loader/pull/469#issuecomment-2625752778 below
(glitchhhhhhhh I can't believe you have hijacked this PR into a successful thing!)


TheGlitch76:
This PR is ready for review now. It allows selecting Mojmap as the runtime namespace in prod, and it uses RuntimeModRemapper to remap all mods in intermediary to mojmap (It successfully loaded the Cobblemon modpack, but might not be fully reliable). Without the experimental system properties, Loader should behave exactly as before.

Some more work will be needed before a hypothetical switch of defaults could occur, at minimum:
- Figuring out how Loader will actually find mojmaps (just download them itself?)
- Figuring out how Mojmap will be named in mod metadata (I'm actually not a fan of the maven identifier, since we're using something that's explicitly not on maven)
- Testing of loading mojmap-compiled mods in a prod mojmap environment

In the interest of getting this merged in a reasonable timeframe, the remaining work will be covered in a different PR.

src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
```diff
@@ -285,7 +285,8 @@ private V1ModMetadataImpl readFields(JsonLoaderValue.ObjectImpl root) {
 			@Nullable
 			JsonLoaderValue intermediateMappingsValue = quiltLoader.get(QLKeys.INTERMEDIATE_MAPPINGS);
 
-			String[] supported_mappings = { "org.quiltmc:hashed", "net.fabricmc:intermediary" };
+			// TODO: "mojang" for mojmap breaks spec; it is just temporary for now because its experimental
+			String[] supported_mappings = { "org.quiltmc:hashed", "net.fabricmc:intermediary", "mojang" };
```
OroArmor: `"net.minecraft:mojang"`?

src/main/java/org/quiltmc/loader/impl/plugin/quilt/StandardQuiltPlugin.java
```diff
@@ -394,9 +394,13 @@ private ModLoadOption[] scan0(Path root, QuiltLoaderIcon fileIcon, ModLocation l
 				context().addFileToScan(inner, jarNode, false);
 			}
 
-			// a mod needs to be remapped if we are in a development environment, and the mod
-			// did not come from the classpath
-			boolean requiresRemap = !location.onClasspath() && QuiltLoader.isDevelopmentEnvironment();
+			// a mod needs to be remapped if the mod did not come from the classpath and its mappings do not match
+			// the current runtime namespace
+			String mappings = meta.intermediateMappings();
```
OroArmor: Id use a switch here

OroArmor (changes requested):

src/main/java/org/quiltmc/loader/impl/plugin/quilt/StandardQuiltPlugin.java
TheGlitch76: this is java 8

OroArmor: you can switch on strings

src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
```diff
@@ -309,6 +310,10 @@ private V1ModMetadataImpl readFields(JsonLoaderValue.ObjectImpl root) {
 				throw new ParseException("Oh no! This version of Quilt Loader doesn't support hashed mappings, please update Quilt Loader to use this mod.");
 			}
 
+			if (mappings.equals("mojang") && !QuiltLoader.getMappingResolver().getNamespaces().contains("mojang")) {
+				throw new ParseException("Oh no! This version of Quilt Loader doesn't support Mojang mappings, please update Quilt Loader to use this mod.");
```
EnnuiL: This copy-pastey comment doesn't make sense here compared to the Hashed Mojmap one; how about:
```suggestion
				throw new ParseException("Oh no! This Quilt Loader installation isn't configured to support Mojang mappings.");
```
(potential remedies to this are to be determined once uhhhh, we figure out a good way to provide mojmap)

src/main/java/org/quiltmc/loader/impl/metadata/qmj/V1ModMetadataReader.java
EnnuiL: glitch says this is temporary, but well? I know how permanent placeholders can be like!
Anyway, `net.minecraft:mojang` makes sense for dev purposes, but Loom uses `net.minecraft:mappings`, which makes sense but is confusing! (what is Minecraft mappings? I've heard of Mojang!)

So, considering this and Mojang's official standards? Maybe we should go with `com.mojang:mappings`?

EnnuiL (commented):
Something I do genuinely wonder about is: what about the mods that uses the remapper thingy? I wouldn't mind breaking compat, but I know that can be tricky; maybe #458 can help?
Anyway, here's the drill:

TheGlitch76:
I've changed the `intermediate_mappings` field to be `org.quiltmc.experimental:mojang` which follows spec. Since this is currently in the same state of usability as plugins (actually, even less so!) I hope this can be an acceptable solution for now.

src/main/java/org/quiltmc/loader/impl/game/MappingConfigurationImpl.java
```diff
@@ -200,9 +204,34 @@ private void initialize() {
...
+		Log.info(LogCategory.MAPPINGS, "Target namespace: %s", getTargetNamespace());
+
+		if (!namespaces.contains(getTargetNamespace())) {
+			if (!namespaces.isEmpty()) {
+				throw new IllegalStateException(String.format("Requested target namespace %s not loaded. Available options: %s", targetNamespace, namespaces));
+			} else {
+				if (getTargetNamespace().equals("official")) {
+					Log.warn(LogCategory.MAPPINGS, "Continuing without mappings because target namespace is 'official'");
+				} else {
+					throw new IllegalStateException("Requested target namespace %s not loaded. To continue without mappings, " +
```
AlexIIL: You either need to use string concatenation here or an explicit `String.format` to handle the `%s` as a target namespace
```suggestion
					throw new IllegalStateException("Requested target namespace " + targetNamespace + " not loaded. To continue without mappings, " +
```

src/main/java/org/quiltmc/loader/impl/transformer/RuntimeModRemapper.java
```diff
@@ -61,28 +69,59 @@ final class RuntimeModRemapper {
...
+			// 	 but wanting to use the default Quilt plugin to change the namespaceMappingFrom
+			//   (in fact, they still have to declare a supported-by-Quilt intermediateMappings in their quilt.mod.json,
+			//   even if their game is not obfuscated at all!)
+			//   Once that is supported, this should change to error or even crash when mods we know are not going to
+			//   be able to load are found.
+
+			Log.info(LogCategory.MOD_REMAP, "Not remapping mods because mappings are empty");
+			return;
+		} else if (!mappingConfiguration.getNamespaces().contains("intermediary")) {
+			Log.warn(LogCategory.MOD_REMAP, "Not remapping mods because intermediary is missing!");
```
AlexIIL, line 88: I think you're missing a `return` here, as this will crash if intermediary is actually missing.

```suggestion
			Log.warn(LogCategory.MOD_REMAP, "Not remapping mods because intermediary is missing!");
			return;
```

src/main/java/org/quiltmc/loader/impl/transformer/KotlinMetadataRemapper.java
```diff
@@ -0,0 +1,80 @@
...
+	// This  is based off the observation that the nice, clean, metadata remapper written for Loom in Kotlin
+	// only uses remapper.map, mapDesc, and mapFieldDesc.
+	// These are extremely identifiable in the metadata's constant pool: field descriptors begin with L,
+	// method descriptors begin with (, and classes we need to remap begin with class_ or contain a package, so we can
+	// just brute-force the constant pool
+	// and remap those 3 things as they come by.
+	@Override
+	public AnnotationVisitor visitAnnotation(String descriptor, boolean visible) {
+		if (descriptor.equals("Lkotlin/Metadata;")) {
+			return new AnnotationNode(QuiltLoaderImpl.ASM_VERSION, descriptor) {
```
AlexIIL: This doesn't send the modified data back to the parent ClassVisitor, you'd need to either add `
accept(KotlinMetadataRemapper.super.visitAnnotation(descriptor, visible));` to both of the returns in `visitEnd`, or replace this with an inline version.

*(I like the inline approach since it avoids building a partial node of a (potentially) long data structure, but just adding the `accept` version twice is the simplest change)*
```suggestion
			return new AnnotationVisitor(api, super.visitAnnotation(descriptor, visible)) {
				@Override
				public AnnotationVisitor visitArray(String name) {
					if ("d2".equals(name)) {
						return new AnnotationVisitor(api, super.visitArray(name)) {
							@Override
							public void visit(String name, Object value) {
								if (value instanceof String) {
									String candidate = (String) value;
									if (candidate.startsWith("(")) {
										candidate = remapper.mapMethodDesc(candidate);
									} else if (candidate.startsWith("L")) { // this could technically catch strays but it should just not do anything
										candidate = remapper.mapDesc(candidate);
									} else if (candidate.startsWith("class_") || candidate.contains("/")) { // must go last to not accidentally catch descriptors
										candidate = remapper.map(candidate);
									} // else hope nothing goes wrong
									value = candidate;
								}

								super.visit(name, value);
							}
						};
					}
					return super.visitArray(name);
				}
			};
		
```

AlexIIL (changes requested):
I've run this against my tests and found a few problems, but otherwise I like the direction!

AlexIIL:
I've pushed all my fixes here: https://github.com/EnnuiL/quilt-loader/pull/1 - feel free to either merge as-is (possibly manually @TheGlitch76 ) or tweak them further?

AlexIIL (approved):
Looks good, thanks!

PR #473, opened 2025-04-08
Fix Mojang mods not running on a Mojang environment

EnnuiL:
This has been tested with a hacked-together mod I made (it's called Ok Zoomer: Mojmap Edition!)

This skips the remapping of Mojmap mods if the environment is Mojang


OroArmor (approved):

AlexIIL:
Yes, this looks good.

There might be a better way of handling this (maybe all mods that have a namespace equal to the target namespace shouldn't be remapped?) But I'm unsure if we need to have all of them for classpath reasons when remapping, so I'll need to look into this more.

PR #476, opened 2025-09-01
Update to Fabric Loader 0.17.2

Codetoil:
I'm not sure if this is enough for a downstreaming update... So I'm setting it up for review.


Codetoil:
Oh wait, I just remembered from earlier... quilt_loader_sat4j (on v2.3.5.1) probably needs to downstream an update from sat4j. Fabric is using sat4j v2.3.6.

src/test/java/net/fabricmc/test/LocalizationTests.java
```diff
@@ -15,14 +15,17 @@
...
 
-	public static <T> void invoke(String name, Class<T> type, Consumer<? super T> invoker) {
-		net.fabricmc.loader.impl.entrypoint.EntrypointUtils.invoke(name, type, invoker);
+public class LocalizationTests {
+	@Test
+	public void formatRoot() {
+		Assertions.assertEquals("client", Localization.format("environment.client"));
+		Assertions.assertEquals("Install A, B.", Localization.format("resolution.solution.addMod", "A", "B"));
 	}
 }
```
AlexIIL, line 31: Please don't delete legacy files - Fabric Loader doesn't have any way to prevent mods from using internal classes, so quilt-loader provides a number of them to ensure old mods still work.

`net.fabricmc.loader.entrypoint.minecraft.hooks.EntrypointUtils` was used by a mod in AOF2 (commit message here: https://github.com/QuiltMC/quilt-loader/blob/58c4cf827703d3bc58c85e0f81a9d78ddf40825e/src/fabric/legacy/java/net/fabricmc/loader/entrypoint/minecraft/hooks/EntrypointUtils.java)

AlexIIL (commented):
This looks good! I still need to test this, but other than this this is a mostly complete update.

I think the only remaining bits are:
- Copying McVersionLookup (as we need that to behave in the same way that fabric-loader does)
- Most likely implementing the field initialization stripping, as fabric mods may now rely on that.

AlexIIL:
Oh, and the sat4j version doesn't need to be bumped - it's an internal library, so mods can't use it. (But there's still no reason why it couldn't be bumped so...)

AlexIIL:
This looks good - thanks!

PR #479, opened 2025-09-09
add system property for enabling the package access hack

SpaceWalkerRS:
this has been in Fabric Loader since 0.16, and at Ornithe we requires it for our new intermediary


MrCat12-Studios (approved):
Seems good to me

AlexIIL:
This looks good, thanks!

AlexIIL:
I will just change the system property check to explicitly require `true` rather than any non-null value though.

PR #480, opened 2025-10-22
Make Quilt Fork address consistent

### LINKED ISSUE #405: Fork comms fails when using IPv6
Sometimes the fork comms client uses IPv6's loopback address (::1), but the server uses IPv4's (127.0.0.1). So the fork comms fails.

pandaninjas:
Resolves #405


AlexIIL:
This looks good, thanks!

PR #493, opened 2026-03-30
support targeting official namespace

eristhea:
patches QL having issues on `26.1`, by only using intermediary when there are actually mappings! made by a friend of mine :)

### DIFF
--- a/src/main/java/org/quiltmc/loader/impl/game/MappingConfigurationImpl.java
+++ b/src/main/java/org/quiltmc/loader/impl/game/MappingConfigurationImpl.java
@@ -68,6 +68,7 @@ public class MappingConfigurationImpl implements MappingConfiguration {
 	private String gameVersion;
 	private String mappingsSource;
 	private VisitableMappingTree mappings = new MemoryMappingTree(true);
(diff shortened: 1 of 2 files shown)


Codetoil:
A proper fix to this would require implementing the custom Mod Distribution Namespaces from fabric, but that would require an API change, and hence a version bump, and should probably come with other changes like a drop in support of CHASM. This will work in the meantime.

AlexIIL:
Thanks for looking into this! However in this case quilt-loader already has support for official mappings - via [`EmptyMappingConfiguration`](https://github.com/QuiltMC/quilt-loader/blob/develop/src/main/java/org/quiltmc/loader/impl/game/EmptyMappingConfiguration.java). So the only thing we need to do is force it to be loaded in [`MinecraftGameProvider`](https://github.com/QuiltMC/quilt-loader/blob/develop/minecraft/src/main/java/org/quiltmc/loader/impl/game/minecraft/MinecraftGameProvider.java#L315-L319).

(The other reason why I'd rather do it this way is to ensure QL still crashes when trying to play an older - obfuscated - versions of minecraft but the mappings file is incorrectly not present)

PR #496, opened 2026-04-30
Fix loader plugins

Codetoil:
Loader Plugins loaded from a jar do not load, instead crashing. This fixes this bug.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,9 +12,9 @@ plugins {
 	id 'java-library'
 	id 'eclipse'
 	id 'maven-publish'
-	id 'org.quiltmc.gradle.licenser' version '2.0.1'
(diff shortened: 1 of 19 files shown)


MrCat12-Studios (changes requested):
Now you need to bump the version to 0.30.0-beta.9 in the `gradle.properties` and `QuiltLoaderImpl.java`.

MrCat12-Studios (changes requested):
You have to bump up the version to 0.30.0-beta.9 in the `gradle.properties` and `QuiltLoaderImpl.java`.

MrCat12-Studios:
Now you have you have to bump the version to 0.30.1-beta.1 or something.

AlexIIL:
Thanks! I'm not sure what I wanted to accomplish with those lines but I'll have to look into them later.

PR #498, opened 2026-06-11
Downstream Fabric Loader 0.19.3

MrCat12-Studios:
I tried my best to downstream the newest fabric loader version. It was relatively small so it didn't take me that many days. I couldn't find any equivalent of [fabric-loader/master:src/test/resources/fabric.mod.json](https://github.com/FabricMC/fabric-loader/blob/master/src/test/resources/fabric.mod.json) with the same tests so I didn't add the tests from fabric's [748f63b](https://github.com/FabricMC/fabric-loader/commit/748f63b3c4c3d1271f4d5b49a59cf2a763a3dfca#diff-9ba34517cab3d734931169f58649ec0bf83d804346f462ae937a1cbb1fa12748). Until last week the file hadn't been updated since 2021 so I am pretty sure there is a reason why they don't exist. I also couldn't find how to implement the changes in [fabric-loader/master:src/main/java/net/fabricmc/loader/impl/FabricLoaderImpl.java](src/main/java/net/fabricmc/loader/impl/FabricLoaderImpl.java) from [a16c6d6](https://github.com/FabricMC/fabric-loader/commit/a16c6d69107a84eee8791a1d35bebcf076b375b7#diff-387fbf9c45c953bd43471185727abc4a1c07d061eab5c0792e15e5d5f1b2b0c6). If anyone could help with that I would be very happy.


AlexIIL:
Thanks for working on this!

PR #499, opened 2026-08-08
Remove the deprecated GUI package in the Plugin API and use the GUI API instead

MrCat12-Studios:
In this PR I have removed completely all uses of the deprecated ...api.plugin.gui and replaced it with it's counterparts in ...api.gui while also adding removing or adapting methods from other classes' exported APIs. Regardless of the PR's ability to be merged, you should keep in mind that removing such largely used API ought to be avoided for minimal updates, and should be kept for a major release. However considering the incredibly small to non-existant amount of quilt mods that do use this API even if they shouldn't it'll probably be fine. Lastly note that maintainers are always free to edit.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,8 +108,8 @@ javaComponent.withVariantsFromConfiguration(configurations.shadowRuntimeElements
 
 test {
 	useJUnitPlatform()
-	testLogging { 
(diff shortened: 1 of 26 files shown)


MrCat12-Studios:
On my computer it built just fine...

MrCat12-Studios:
Great now it works

PR #503, opened 2026-08-27
Remove shell parameter-gating loader plugins

Olafcio1:
pwease

PR #505, opened 2026-08-27
Loader plugins cleanup

AlexIIL:
I still need to RFC loader plugins, but they are basically not going to change after this, so there's little reason to prevent people from using them right now.

This superseeds #503, and should release properly long before we remove the deprecated `plugins.gui` package in #499

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,6 +105,9 @@ javaComponent.withVariantsFromConfiguration(configurations.shadowRuntimeElements
 
 test {
 	useJUnitPlatform()
+	testLogging {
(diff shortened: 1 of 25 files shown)


MrCat12-Studios:
Should you draft it?

AlexIIL:
No, I want to push this out as a beta either tomorrow or at the weekend, since we have an ~~impatient~~ *excited* modder who wants to try using loader plugins in a production environment.

MrCat12-Studios:
Lol

MrCat12-Studios:
What about adding it as a snapshot? I get it, there hasn't been one since 0.18 or whatever but I think it's better if we add a 0.31 pre release s snapshot rather than a 0.30.1 beta

